PR #167, opened 2024-06-20
migrate ResearchGroup to Project

lukavdplas:
This migrates existing `ResearchGroup` data into the `Project` model.

- All `ResearchGroup` objects are saved as a `Project` model.
- For collections and annotation, adds a `context` field to replace `managing_group`.

The intention is that after this migration, the `ResearchGroup` model can be worked out of the application (and replaced with `Project`). I'll add a PR for that in a bit.


backend/projects/migration_utils.py
```diff
@@ -0,0 +1,7 @@
+import re
+
+def name_to_slug(name: str, max_length: int = 64) -> str:
+    lowercase = name.lower()
+    no_whitespace = re.sub(r'\s', '_', lowercase)
+    cleaned = re.sub(r'[^a-z_]', '', no_whitespace)
+    return cleaned[:min(max_length, len(cleaned))]
```
tijmenbaarda, line 7: This function will yield empty strings if input consists of non-Latin characters only. But this function will only be used for a one-off migration, correct?

tijmenbaarda (approved):
Very nice, especially that the migration works in two ways!

How have you tested this with real data? Maybe I can test next week with the data from the production server (on a local installation, of course), if you haven't done so.

backend/projects/migration_utils.py
lukavdplas: Yes, it's only used here. In new models, the URI will be based on the `name`, which is a `SlugField` that can't be blank.

lukavdplas:
>How have you tested this with real data? Maybe I can test next week with the data from the production server (on a local installation, of course), if you haven't done so.

No, only on my local dataset.

lukavdplas:
Note: I've updated the URI format for projects to be consistent with #181. If you had already run the migration here, revert it and run it again.

PR #168, opened 2024-06-20
Feature/researchgroup to project

lukavdplas:
This PR should complete the migration to a new `Project` model. Builds on #167 

It updates all functionality based on `ResearchGroup` by referring to `Project` instead. There should not be any noticable changes to users in the frontend.

Also adds a migration to delete the `ResearchGroup` model.


backend/vre/models.py
```diff
@@ -58,9 +40,8 @@ class Annotation(models.Model):
     An annotation is also linked to exactly one research group,
```
tijmenbaarda, line 40: Update docstring?

tijmenbaarda (approved):
Looks good, one minor comment!

PR #170, opened 2024-06-25
RDF data models

lukavdplas:
This is some preparation I did for managing collections and annotations. It adds classes for modelling RDF data that feel somewhat similar to Django's model layer. See the changes in the `project` app for an example.

I did not want to spend too much time on this right now; the current state is quite basic, but I think it's developed enough that it will help with writing the backend for collections + annotations.


backend/triplestore/rdf_model.py
```diff
@@ -0,0 +1,90 @@
...
+from django.conf import settings
+
+from triplestore.rdf_field import RDFField
+from triplestore.utils import triples_to_quads, Triples
+
+class RDFModel():
+    '''
+    Abstract class for RDF data models
+
+    Create subclasses to model a data class.
```
tijmenbaarda, line 16: Also indicate how to do this: by optionally setting `rdf_class` and adding `RDFProperty`s.

backend/triplestore/rdf_field.py
```diff
@@ -0,0 +1,246 @@
...
+            self.node_to_value(o)
+            for (s, p, o) in self._stored_triples(instance)
+        ]
+
+    def node_to_value(self, node: Identifier):
+        '''
+        Convert an object node to its modelled value
+        '''
+        if isinstance(node, Literal):
+            return node.value
```
tijmenbaarda, line 157: I understand how this works if `node` is a `Literal`, but what about other cases? Are these supported at all?

backend/triplestore/rdf_field.py
```diff
@@ -0,0 +1,246 @@
...
+        return node
+
+    def value_to_node(self, value) -> Identifier:
+        '''
+        Convert a modelled value to an object identifier.
+        '''
+        if isinstance(value, Identifier):
+            return value
+        else:
+            return Literal(value)
```
tijmenbaarda, line 167: Will this work with other data types than strings?

backend/triplestore/rdf_field.py
```diff
@@ -0,0 +1,246 @@
...
+    def _triples_to_store(self, instance, value) -> utils.Triples:
+        '''
+        Triples that should be stored to represent the modelled value
+        '''
+        raise NotImplementedError()
+
+
+class RDFPredicateField(RDFField):
+    '''
+    Models a predicate, i.e. a property + object in the graph. This translates to a
```
tijmenbaarda, line 105: A predicate in RDF is the second part of the triple, not the second and the third part together, I think?

backend/triplestore/rdf_field.py
```diff
@@ -0,0 +1,246 @@
...
+    boolean value in the model.
```
tijmenbaarda, line 106: Not very clear to me - perhaps: the `get` method returns a boolean value that is true if the triple exists in the graph, otherwise false? (If that is what you mean)

backend/triplestore/rdf_field.py
```diff
@@ -0,0 +1,246 @@
...
+    fetch the correct graph for reading and writing triples. By default, this will be the
+    graph of the model instance. To use a constant graph, you can pass a `graph` in the
+    initialiser. If the graph is chosen dynamically, you can override the method
+    `get_graph`.
+    '''
+
+    def __init__(self, graph: Optional[Graph] = None, **kwargs):
+        self.graph = graph
+
+    def get_graph(self, instance):
```
tijmenbaarda, line 42: Here and at other places it was not very clear to me immediately what is expected for `instance`. And why is that necessary, if the graph may also be stored in the object during initialization?

tijmenbaarda (commented):
I think this is really cool! A really nice and clean Django-style solution.

We see the framework in action here in a case where the properties are all (string?) literals. Will this also work with URIs and blank nodes? I would suggest adding some documentation about that and mentioning what can be expected from the framework and what not (for example, the properties of the subject itself will be read, added, updated and deleted, but any nested data will be left untouched -- which I think will be fine, because this framework will only be used with data that is under the control of the application, I would say).

backend/triplestore/rdf_field.py
lukavdplas: I've added some clarification to the docstring. If the node is a `BNode` or `URIRef`, it will be returned as-is. Alternatively, you could create a subclass that handles values some other way.

PR #175, opened 2024-06-27
Add missing trailing commas to API URLs

jgonggrijp:
Without these changes, this trigger:

https://github.com/UUDigitalHumanitieslab/EDPOP/blob/1f5fe3497fc35f24c819122da0a3a6c98e61514d/frontend/vre/main.js#L112

never happens for me, and the collection dropdown is not inserted in the DOM as a result. This is because Django does not recognize the URLs without a trailing slash and defaults to the index page. The collections in question cannot read the HTML and therefore never trigger their `update` events, causing the `finish` trigger to fail.

@tijmenbaarda It is unclear to me why this does not seem to have affected you. Please check whether the application still works for you with these changes. Manually load http://localhost:8000/static/bundle.js first and refresh to make 100% sure that the browser is running the latest code.

### DIFF
--- a/frontend/vre/collection/collection.model.js
+++ b/frontend/vre/collection/collection.model.js
@@ -23,7 +23,7 @@ export var VRECollection = APIModel.extend({
 });
 
 export var VRECollections = APICollection.extend({
-    url: '/api/collections',
+    url: '/api/collections/',
     model: VRECollection,
 }, {
     /**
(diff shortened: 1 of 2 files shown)


tijmenbaarda:
I have no idea why things were working on my installation, but with this branch it still works and also my code in #174 still works when merging this branch into it.

PR #176, opened 2024-06-28
Add BSD-3-Clause license

jgonggrijp:
With the source code going public, it is time to add a license. I stop for a review because I am not 100% sure that this license was the plan.


tijmenbaarda:
Seems good to me!

Issue #131, opened 2023-01-16
Render-placement antipattern


jgonggrijp:
After #130, this "gotcha" seems worthwhile to document as well.

The following pattern frequently occurs in snippets of Backbone example code, scattered across the web:

```js
someView.render().$el.appendTo(someTargetElement);
```

`someView.render()` returns `someView` by convention, so the above pattern is equivalent to the following two lines, which by themselves are both perfectly reasonable:

```js
// rendering the view
someView.render();

// placing the view
someView.$el.appendTo(someTargetElement);
```

The single-line version, where rendering and placing the view happen in one go, frequently appears in example code because it is a convenient, simple way to demonstrate features of the library. However, it is a form of code smell when encountered in production code. (Side note: other libraries have similar problems with example code that should not be imitated too literally.) While it is safe to combine the operations, they generally happen for very different reasons, which tend to be separated both in execution time and code space, so seeing them in the same place is a likely indicator of improper state management.

*Rendering* a view happens in order to align its contents with whatever it is meant to present. It only affects the internal structure of the view's DOM element. For this reason, it is generally also the responsibility of the view to decide when it will render *itself*, for example when the underlying model changes. This can be done regardless of whether the view is visible to the user at that time. The `initialize` method of a view commonly contains `this.render()` and/or something like `this.listenTo(this.model, 'change', this.render)`. In fact, the only reason *not* to self-render a view on creation is when rendering is too expensive, but this in itself is a situation that should be avoided when possible.

*Placing* a view happens in order to integrate it in a larger context and make it visible to the user. This never affects the view's internal structure and the view itself generally has not enough information to decide where it should be placed, or when. One could say that placement is the external complement to rendering.

Issue #130, opened 2023-01-16
Some frontend views are leaky


jgonggrijp:
This is a relatively minor code quality issue, but it still deserves to be documented in a ticket.

Whenever event handlers are registered with methods like `.on`, `.once`, `.listenTo` et al, this introduces a potential memory leak: if the event emitter is `delete`d but the listener is not, the reference from the listener to the emitter prevents the latter from being purged from memory. This is especially expensive when the emitter is a DOM element. For this reason, `Backbone.View` comes with a `remove` method, which by default cuts all ties between the view and its model or collection, as well as between the view and its DOM element (`view.el`/`view.$el`).

In some cases, a view will introduce "special" references, for example when attaching `"click"` event handlers directly to `document.body` in order to implement "click outside to hide" behavior, or when introducing dynamic elements through Bootstrap's jQuery plugins. In such cases, the view *should* extend the `remove` method in order to clean up those references (by removing the external event handlers, destroying the instance of the jQuery plugin, etcetera). I have seen one view where such special references were created in the `initialize` method, but the `remove` method was not extended accordingly (to be referenced below); there might be more.

Another common way in which memory leaks can be introduced in Backbone applications, is when the `remove` method is never called, even though a view is no longer used. This can happen when a view has subviews that are managed manually, which is currently the case for several views in the EDPOP VRE, and the lifetime of the subviews is improperly managed. In most cases, this problem disappears automatically when composition is handed off to backbone-fractal (#123).

I suggest that we simply address instances of this issue when we encounter them and reference this issue when we do so (to be exemplified below). At some point, we *might* want to sweep up the remaining instances in one go, so we can close this ticket. *If* we do so, this would be best done after #123.

Issue #123, opened 2022-09-22
Use backbone-fractal and backbone.radio where applicable


jgonggrijp:
Apply the lessons learned in READ-IT.

PR #178, opened 2024-07-11
Reactivity

### LINKED ISSUE #177: Use a model to track the currently selected component
See the end of the discussion in #174

jgonggrijp:
This branch closes #177 and also partially addresses #123, #130 and #131.

These changes demonstrate a few tricks and principles (apart from a few other, minor cleanups):

- View composition. Nested views are owned by their enveloping views, rather than being injected from the outside by unrelated code.
- Views decide for themselves when to render.
- Sharing models and collections between views whenever possible.
- Long distance communication between views through events in shared models, rather than through global state. For example, each entry in the catalog dropdown and the collection dropdown is now a separate view that is aware of the catalog/collection it represents. It will update its `active` class in response to `focus`/`blur` events from its `.model` (i.e., the catalog or collection in question). This avoids having to redraw both dropdown menus entirely every time the selection changes.
- Using `enter` and `exit` events on a state model, to insert a new view into the DOM and also cleanly remove it again when appropriate. This is also used to trigger the `focus`/`blur` events on the catalogs and collections.
- Using `route` events from a router, rather than defining route-handling methods directly within the router itself.
- Loose coupling through reactivity (also known as the law of Demeter): individual model and view classes know relatively little about each other, as can be seen from the reduced usage of `GlobalVariables`. The main.js creates instances of these classes and adds coupling from the outside, mostly by binding methods of one instance to events of another.

I write "demonstrate", because there is also still some code that could use the above treatment, but which I skipped. That means that #123, #130 and #131 cannot be closed yet.

You are cordially invited to read the code changes for inspiration. I expect at least some of the changes to raise questions; please do not hesitate to ask for clarification where needed.

I expect these changes to conflict with changes that you have written in parallel. I propose to merge your changes first, I will rebase my changes on yours after that.

### DIFF
--- a/backend/catalogs/api.py
+++ b/backend/catalogs/api.py
@@ -20,17 +20,17 @@ class SearchView(RDFView):
     
     def get_graph(self, request: views.Request, **kwargs) -> Graph:
         try:
-            catalog = request.query_params["catalog"]
+            source = request.query_params["source"]
             query = request.query_params["query"]
             start = request.query_params.get("start", "0")
             end = request.query_params.get("end", None)
         except KeyError as err:
             raise ParseError(f"Query parameter missing: {err}")
-        assert isinstance(catalog, str)
+        assert isinstance(source, str)
         assert isinstance(query, str)
         assert isinstance(start, str)
         assert end is None or isinstance(end, str)
-        catalog_uriref = URIRef(catalog)
+        catalog_uriref = URIRef(source)
         start = int(start)
         if end is not None:
             end = int(end)
--- a/frontend/Dockerfile
+++ b/frontend/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:14
+FROM node:16
 
 WORKDIR /usr/src/app
 COPY package.json package-lock.json .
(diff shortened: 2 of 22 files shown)


tijmenbaarda (approved):
Great!

I must admit that it took me some time to understand how this works. Part of that was because it is hard to find out what kind of values are expected in the various properties of the `navigationState` model. Is there a usual way to document what the model's properties mean?

I think also `collection.search.view.js` has to be renamed to `catalog.search.view.js`, but maybe it is better to do that after #183 is merged.

jgonggrijp:
> I must admit that it took me some time to understand how this works.

No shame in that, especially given that it was already expected.

> Part of that was because it is hard to find out what kind of values are expected in the various properties of the `navigationState` model. Is there a usual way to document what the model's properties mean?

Not that I'm aware of, but we can probably agree on something. Do you have suggestions?

> I think also `collection.search.view.js` has to be renamed to `catalog.search.view.js`, but maybe it is better to do that after #183 is merged.

Agreed!

tijmenbaarda:
@jgonggrijp : I have merged develop into your branch so that I can continue with the next step. The merge conflicts were not dramatic and it helped me to understand your new set-up better.

jgonggrijp:
@tijmenbaarda I will rebase this branch on `develop` before merging, because the history will be cleaner in that way. However, it is no problem if you already based new work on this branch; you can rebase your new branch on `develop` again afterwards.

PR #181, opened 2024-07-16
Feature/collection api

### LINKED ISSUE #159: Collection list endpoint
This is an RDF endpoint (#157) where clients should be able to

- [ ] Send a GET request to get a list of the collections that exist (sourced from the `edpop:collections` graph, see #156).
- [ ] Send a POST request to create a new collection.
    - [ ] The creator determines the IRI of the collection, which ends in a short name. For example `edpop:collection/almanacs`. This IRI is already used in the request payload, so there is no need to work with blank nodes.
    - [ ] The request fails if a collection with that IRI already exists.
    - [ ] Creating a new collection should specify to which usergroup the collection belongs. The user must already be a member of that research group.
    - [ ] Possible alternative: create a collection that does not belong to a research group but which is private to the user (though maybe still visible for other users).

### LINKED ISSUE #160: Single collection endpoint
This is an RDF endpoint (#157) where a client can

- [ ] Send a GET request to retrieve information about a single collection.
    - This does **not** include a listing of records within the collection.
    - This does **not** include a listing of users that have access to the collection.
- [ ] Send a PUT or PATCH request to change the collection (primarily its name).
- [ ] Send a DELETE request in order to delete the collection.
- [ ] Send POST requests with varying payloads in order to add and remove records to/from the collection.
    - [ ] These operations require a small domain-specific language (DSL) which could be based either on RDF or plain JSON.

The modifying operations (everything except for the GET) must be only possible if the user has sufficient permission to execute them.

Best implemented *after* #159.

lukavdplas:
This adds an API for collections which are saved in the triplestore. close #159, close #160

It adds list, retrieve, create, update, and delete endpoints for collections. (Which replace the existing API for collections.)

This is implemented as a "regular" JSON API, but collections are identified by an URI instead of an integer.

This update only implements the API on the backend, so it will break the frontend (which expects a slightly different API for collections). I expect adjusting the frontend will go smoothly, but I did not want to bloat this PR.

In the meantime, you can test the browsable API at `/api/collections/`.

This update also does not include a data migration or adjust the existing data migration script; I'll add that in a later PR.

Other changes:

- Projects now explicitly store their URI in the database. This makes lookups easier.
- The core URL configuration is moved from the `vre` app to the project module `edpop`.
- Adds defined namespaces (https://github.com/UUDigitalHumanitieslab/EDPOP/pull/180)


backend/collect/utils.py
```diff
@@ -0,0 +1,29 @@
...
+    clean = lambda s: re.sub(r'[^a-z0-9\-_]', '', s)
+    strip_start = lambda s: re.sub(r'^\W+', '', s)
+    strip_end = lambda s: re.sub(r'\W+$', '', s)
+
+    return strip_end(strip_start(clean(no_spaces(str.lower(name)))))
+
+
+def collection_uri(name: str):
+    id = _name_to_slug(name)
+    return URIRef(settings.RDF_NAMESPACE_ROOT + 'collections/' + id)
```
lukavdplas, line 18: This constructs the URI for a collection based on the "human" name, i.e. "My collection" will be stored as `my_collection`. This simplifies the frontend UI, which is nice for an initial version, though it may lead to some annoyances long-term.

jgonggrijp: What annoyances?

lukavdplas: URIs need to be unique, so the request will fail if the URI already exists. This isn't really a problem right now, but if a group of people is enthusiastically adding data on the same topics, conflicts may become more common.

I think there are some minor annoyances that can pop up if you get these warnings:

- You can get a response that the URI is taken while the existing collection is privately owned by another user. This makes technical sense, but it feels odd.
- Names don't need to be unique but URIs do, so this can feel un-intuitive. For instance, if "My collection" / `my_collection` exists, you can get the name you want by creating "My collection 2" / `my_collection_2`, and then renaming it to "My collection" (which will preserve the URI `my_collection_2`).
- "my collection" and "My Collection :)" would both be encoded as `my_collection`, so the user may feel like they don't have full control over the URI. Which is annoying if their collection gets rejected on the basis of the URI.

jgonggrijp:
@lukavdplas I merged `develop` into a local copy of your branch so I could use Docker to try the browsable API. There were a few minor merge conflicts that were easy to fix and I had to update `backend/requirements.{in,txt}` to replace the `feature/json-ld` branch by `develop` for the `restframework_rdf` package, because the former branch was already merged.

Unfortunately, I cannot get past the migrations of the projects app. It gets stuck on number 2, log below. Does this look familiar to you and if so, do you know how to address it? The `Project` class does have a `graph` method.

```
2024-07-17 17:14:53 System check identified no issues (0 silenced).
2024-07-17 17:14:54 Operations to perform:
2024-07-17 17:14:54   Apply all migrations: admin, auth, authtoken, contenttypes, projects, sessions, vre
2024-07-17 17:14:54 Running migrations:
2024-07-17 17:14:54 Traceback (most recent call last):
2024-07-17 17:14:54   File "/usr/src/app/backend/manage.py", line 22, in <module>
2024-07-17 17:14:54     execute_from_command_line(sys.argv)
2024-07-17 17:14:54   File "/usr/local/lib/python3.9/site-packages/django/core/management/__init__.py", line 442, in execute_from_command_line
2024-07-17 17:14:54     utility.execute()
2024-07-17 17:14:54   File "/usr/local/lib/python3.9/site-packages/django/core/management/__init__.py", line 436, in execute
2024-07-17 17:14:54     self.fetch_command(subcommand).run_from_argv(self.argv)
2024-07-17 17:14:54   File "/usr/local/lib/python3.9/site-packages/django/core/management/base.py", line 412, in run_from_argv
2024-07-17 17:14:54     self.execute(*args, **cmd_options)
2024-07-17 17:14:54   File "/usr/local/lib/python3.9/site-packages/django/core/management/base.py", line 458, in execute
2024-07-17 17:14:54     output = self.handle(*args, **options)
2024-07-17 17:14:54   File "/usr/local/lib/python3.9/site-packages/django/core/management/base.py", line 106, in wrapper
2024-07-17 17:14:54     res = handle_func(*args, **kwargs)
2024-07-17 17:14:54   File "/usr/local/lib/python3.9/site-packages/django/core/management/commands/migrate.py", line 356, in handle
2024-07-17 17:14:54     post_migrate_state = executor.migrate(
2024-07-17 17:14:54   File "/usr/local/lib/python3.9/site-packages/django/db/migrations/executor.py", line 135, in migrate
2024-07-17 17:14:54     state = self._migrate_all_forwards(
2024-07-17 17:14:54   File "/usr/local/lib/python3.9/site-packages/django/db/migrations/executor.py", line 167, in _migrate_all_forwards
2024-07-17 17:14:54     state = self.apply_migration(
2024-07-17 17:14:54   File "/usr/local/lib/python3.9/site-packages/django/db/migrations/executor.py", line 252, in apply_migration
2024-07-17 17:14:54     state = migration.apply(state, schema_editor)
2024-07-17 17:14:54   File "/usr/local/lib/python3.9/site-packages/django/db/migrations/migration.py", line 132, in apply
2024-07-17 17:14:54     operation.database_forwards(
2024-07-17 17:14:54   File "/usr/local/lib/python3.9/site-packages/django/db/migrations/operations/special.py", line 193, in database_forwards
2024-07-17 17:14:54     self.code(from_state.apps, schema_editor)
2024-07-17 17:14:54   File "/usr/src/app/backend/projects/migrations/0002_researchgroups_to_projects.py", line 22, in research_groups_to_projects
2024-07-17 17:14:54     store_project_graph(Project, project, True)
2024-07-17 17:14:54   File "/usr/src/app/backend/projects/signals.py", line 21, in store_project_graph
2024-07-17 17:14:54     g = instance.graph()
2024-07-17 17:14:54 AttributeError: 'Project' object has no attribute 'graph'
2024-07-17 17:14:54   Applying projects.0002_researchgroups_to_projects...
```

lukavdplas:
It does. This happens because migrations use a reconstruction of the model based on the migration history, rather than the actual class. So this migration tried to access the custom method `graph()`, but that isn't available during the migration.

This is caused by a mistake in the source code, I'll update it and let you know.

lukavdplas:
Okay, I've updated it and it looks like the migrations run fine now. I'm not very happy with the solution (importing the actual model class during the migration) but it works.

backend/collect/utils.py
jgonggrijp: Maybe number slugs automatically, then?

jgonggrijp:
Getting one for `Collection` as well:

```
2024-07-18 14:48:00 System check identified no issues (0 silenced).
2024-07-18 14:48:00 Operations to perform:
2024-07-18 14:48:00   Apply all migrations: admin, auth, authtoken, contenttypes, projects, sessions, vre
2024-07-18 14:48:00 Running migrations:
2024-07-18 14:48:01   Applying projects.0002_researchgroups_to_projects... OK
2024-07-18 14:48:01   Applying projects.0003_project_uri... OK
2024-07-18 14:48:01   Applying projects.0004_fill_project_uri... OK
2024-07-18 14:48:01   Applying projects.0005_alter_project_uri... OK
2024-07-18 14:48:01 Traceback (most recent call last):
2024-07-18 14:48:01   File "/usr/src/app/backend/manage.py", line 22, in <module>
2024-07-18 14:48:01     execute_from_command_line(sys.argv)
2024-07-18 14:48:01   File "/usr/local/lib/python3.9/site-packages/django/core/management/__init__.py", line 442, in execute_from_command_line
2024-07-18 14:48:01     utility.execute()
2024-07-18 14:48:01   File "/usr/local/lib/python3.9/site-packages/django/core/management/__init__.py", line 436, in execute
2024-07-18 14:48:01     self.fetch_command(subcommand).run_from_argv(self.argv)
2024-07-18 14:48:01   File "/usr/local/lib/python3.9/site-packages/django/core/management/base.py", line 412, in run_from_argv
2024-07-18 14:48:01     self.execute(*args, **cmd_options)
2024-07-18 14:48:01   File "/usr/local/lib/python3.9/site-packages/django/core/management/base.py", line 458, in execute
2024-07-18 14:48:01     output = self.handle(*args, **options)
2024-07-18 14:48:01   File "/usr/local/lib/python3.9/site-packages/django/core/management/base.py", line 106, in wrapper
2024-07-18 14:48:01     res = handle_func(*args, **kwargs)
2024-07-18 14:48:01   File "/usr/local/lib/python3.9/site-packages/django/core/management/commands/migrate.py", line 356, in handle
2024-07-18 14:48:01     post_migrate_state = executor.migrate(
2024-07-18 14:48:01   File "/usr/local/lib/python3.9/site-packages/django/db/migrations/executor.py", line 135, in migrate
2024-07-18 14:48:01     state = self._migrate_all_forwards(
2024-07-18 14:48:01   File "/usr/local/lib/python3.9/site-packages/django/db/migrations/executor.py", line 167, in _migrate_all_forwards
2024-07-18 14:48:01     state = self.apply_migration(
2024-07-18 14:48:01   File "/usr/local/lib/python3.9/site-packages/django/db/migrations/executor.py", line 252, in apply_migration
2024-07-18 14:48:01     state = migration.apply(state, schema_editor)
2024-07-18 14:48:01   File "/usr/local/lib/python3.9/site-packages/django/db/migrations/migration.py", line 132, in apply
2024-07-18 14:48:01     operation.database_forwards(
2024-07-18 14:48:01   File "/usr/local/lib/python3.9/site-packages/django/db/migrations/operations/special.py", line 193, in database_forwards
2024-07-18 14:48:01     self.code(from_state.apps, schema_editor)
2024-07-18 14:48:01   File "/usr/src/app/backend/vre/migrations/0006_annotation_context_collection_context.py", line 46, in save_collection_managing_group_as_context
2024-07-18 14:48:01     matching_name = obj.managing_group.filter(name=obj.name)
2024-07-18 14:48:01 AttributeError: 'Collection' object has no attribute 'name'
2024-07-18 14:48:01   Applying vre.0006_annotation_context_collection_context...
```

lukavdplas:
Ah, that should be fixed now! Thanks for trying all this out btw 😅

backend/projects/rdf_models.py
```diff
@@ -1,9 +1,45 @@
...
+from django.conf import settings
+
 from triplestore.constants import AS, EDPOPCOL
 from triplestore.rdf_model import RDFModel
-from triplestore.rdf_field import RDFUniquePropertyField
+from triplestore.rdf_field import RDFUniquePropertyField, RDFQuadField
+from triplestore.utils import Quads
+
+
+class CollectionsField(RDFQuadField):
```
jgonggrijp, line 11: Please add a docstring to explain the purpose of this class.

backend/projects/rdf_models.py
```diff
@@ -1,9 +1,45 @@
...
+        SELECT ?col WHERE {{                              
+                ?col a edpopcol:Collection ;
+                as:context <{instance.uri}> .
+        }}
+        ''', initNs={'as': AS, 'edpopcol': EDPOPCOL})
+
+        return [
+            (result, AS.context, instance.uri, Graph(store, result))
+            for (result, ) in results
+        ]
```
jgonggrijp, line 30: I am not sure, but it looks to me as if the query in `_stored_quads` gets the information that `get` is looking for. However, `_stored_quads` decorates this information with additional data, which `get` then peels off again. Wouldn't it be more straightforward and efficient to put the query in `get` and then call `get` from `_stored_quads`?

backend/projects/signals.py
```diff
@@ -1,9 +1,18 @@
...
 from projects.models import Project
 from projects.rdf_models import RDFProject
 
+@receiver(pre_save, sender=Project)
+def set_project_uri(sender, instance: Project, **kwargs):
+    '''
+    Set project URI if it is empty.
+    '''
+    if not instance.uri:
+        instance.uri = settings.RDF_NAMESPACE_ROOT + 'projects/' + instance.name
```
jgonggrijp, line 14: Here's a second place where you define the URI of a project. This one is probably better. Regardless, you should factor out the logic to a single source of truth.

jgonggrijp (changes requested):
I was a bit confused by this comment:

> Current status: this PR is ready for review; I've made it a draft because it should only be merged in combination with a frontend update.

The pull request compares to `feature/rdf-modeling-utils`, not `develop`. Surely, you are planning to build the `feature/rdf-modeling-utils` branch incrementally, and the current branch does not need to be complete with frontend adaptations before you merge it?

Side note: even on `develop`, I wouldn't necessarily be against you merging this without corresponding frontend changes. We just did something like that in #174. As far as I'm concerned, only on `master`/`main` does everything need to be completely consistent and bug-free. The purpose of release branches is to ensure this.

Anyway, back to the actual changes. I like the cleanness of the code, but I would prefer more docstrings and comments to explain why things are the way they are. I also noticed a bug.

I created a new collection with the slug `extra` through the browsable API. The URI of this collection is `http://localhost:8000/rdf/collections/extra`. Create, retrieve, update and delete all work. However:

- In the triplestore, the slug of the associated project is duplicated. Its URI is `http://localhost:8000/rdf/project/Dev/Dev` everywhere (so at least the URI is consistent). Note that `project` is singular as well. By contrast, my project that was migrated from a group has a plural `projects` and contains the slug only once: `http://localhost:8000/rdf/projects/dev_g`.
- I updated my newly created collection to associate it with the `dev_g` project instead. However, the corresponding object node in the triplestore now held the URI `http://localhost:8000/rdf/project/dev_g/dev_g`.

Some more details in the comments. Keep up the sweet refactoring!

backend/projects/rdf_models.py
lukavdplas: I get the confusion. The reason is that `_stored_quads` is also used to update the triplestore. When you call `save()` on the model, the field will compare `_stored_quads()` with `_quads_to_store()` to update the graph. `_stored_quads()` is also used when you call `delete()`.

So `_stored_quads()` returns the representation that is saved in the triplestore, which is needed to make updates. `get()` translates the graph representation to one that is more convenient in the model.

This is implemented on the parent class (`RDFQuadField`), but it looks convoluted in isolation.

backend/collect/api_test.py
```diff
@@ -0,0 +1,115 @@
...
+
+
+def test_create_fails_if_collection_exists(db, user, project, client: Client):
+    client.force_login(user)
+    success_response = post_collection(client, project.name)
+    assert is_success(success_response.status_code)
+
+    # try to create the same collection again
+    fail_response = post_collection(client, project.name)
+    assert is_client_error(fail_response.status_code)
```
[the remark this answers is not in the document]
jgonggrijp: You are right about Blazegraph, but I meant the representation of the collection in the PostgreSQL database. Hence "once database synchronization is added".

backend/projects/rdf_models.py
jgonggrijp: Sure, but you are in full control here because you are overriding both methods. What stops you from writing this?

```suggestion
    def get(self, instance: RDFModel):
        store = settings.RDFLIB_STORE
        results = store.query(f'''
        SELECT ?col WHERE {{                              
                ?col a edpopcol:Collection ;
                as:context <{instance.uri}> .
        }}
        ''', initNs={'as': AS, 'edpopcol': EDPOPCOL})
        return [
            s
            for (s, ) in results
        ]

    def _stored_quads(self, instance: RDFModel) -> Quads:
        store = settings.RDFLIB_STORE
        
        return [
            (result, AS.context, instance.uri, Graph(store, result))
            for result in self.get(instance)
        ]
```

lukavdplas: Ah right. Nothing, but to explain why it wasn't written like that: in general, `_stored_triples`/`_stored_quads` may contain information that is ignored by the model (and thus not retrievable from the output of `get()`).

For instance, you could write this field to be agnostic about the graph context. In that case, `_stored_quads` should include the graph the `as:context` triple was stored in, because that information is necessary for save/delete operations. But `get()` can filter it out.

So *in general*, it makes sense to write `_stored_triples`/`_stored_quads` first, and then convert that to an internal value in `get()`. In this specific case, though, you're right that the dependency can flow either way.

jgonggrijp:
@lukavdplas Is this ready to merge? Should I have another look at it first?

backend/collect/api_test.py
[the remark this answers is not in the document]
lukavdplas: I see! Though it was my understanding that we plan to only save collections in blazegraph in the future.

lukavdplas:
I think it's ready to merge. Bear in mind that this API isn't compatible with the frontend, so it will make the develop branch not-release-ready for a while. But I think we already discussed that's not an issue.

backend/collect/api_test.py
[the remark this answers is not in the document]
jgonggrijp: It depends on how we administrate which people have (write) access to which collections.

Issue #182, opened 2024-07-16
Correctly handle duplicate records in external catalogs


tijmenbaarda:
Some external APIs (VD17) sometimes return multiple identical records, which also have the same ID. After converting to linked data, the duplication is lost because the subjects for duplicate records are the same. However, the frontend still expects to receive the requested number of records, and fields are duplicated because they are created every time by the Explorer.

tijmenbaarda:
This has been solved in EDPOP Explorer, so this issue can be closed.

PR #183, opened 2024-07-16
Basic catalog searching

tijmenbaarda:
Allow searching using the new catalog system, via an RDF representation in JSON-LD. Showing record details is not yet working, I am working on that in a separate branch. This branch contains the code to make sense of JSON-LD responses with nested information and with a specified order of subjects using the ActivityStreams scheme.

Known issue: #182 (Correctly handle duplicate records in external catalogs)


frontend/vre/utils/jsonld.model.js
```diff
@@ -1,11 +1,91 @@
...
+ * @returns {Object}
+ */
+export function nestSubject(graph, subject) {
+    const recursiveFunction = function(graph, subject, baseSubject) {
+        const transformedSubject = _.clone(subject);
+        for (let property of Object.keys(subject)) {
+            if (subject[property].hasOwnProperty("@id")) {
+                // This is a reference to another subject
+                const refereedSubject = graph.find((thisSubject) => thisSubject["@id"] === subject[property]["@id"]);
+                if (refereedSubject && refereedSubject["@id"] !== baseSubject["@id"]) {
```
jgonggrijp, line 41: I don't know what scenarios are realistic here, but in the most general case, checking only against `baseSubject` is not enough. Cycle detection in recursive algorithms generally requires tracking the full recursion stack, because a cycle need not necessarily involve the root entry.

I will suggest some test code that demonstrates the theoretical problem in a later comment, but feel free to ignore it if you think that the current code covers all realistic scenarios.

frontend/vre/utils/jsonld.model.js
```diff
@@ -1,11 +1,91 @@
...
+                       Only do this if the subject is not the same as the one we started with,
+                       to avoid an endless loop. (Alternative would be to create a circular reference) */
+                    transformedSubject[property] = recursiveFunction(graph, refereedSubject, baseSubject);
+                }
+            }
+        }
+        return transformedSubject;
+    }
+    return recursiveFunction(graph, subject, subject);
+}
```
jgonggrijp, line 52: Despite all the comments about this function, I should add that I still agree with the general approach. In fact, I like the brevity.

frontend/vre/utils/jsonld.model.js
```diff
@@ -1,11 +1,91 @@
...
+     */
+    activityStreamsPrefix: "as:",
+    parse: function(response) {
+        // Get all subjects of the graph with their predicates and objects as an array
+        if (!response.hasOwnProperty("@graph")) {
+            throw "Response has no @graph key";
+        }
+        const allSubjects = response["@graph"];
+        const orderedCollection = allSubjects.find((subject) => {return subject["@type"] === `${this.activityStreamsPrefix}OrderedCollection`}, this);
+        this.totalResults = orderedCollection[`${this.activityStreamsPrefix}totalItems`] ?? null;
```
jgonggrijp, line 82: Is it important that `this.totalResults` is `null` rather than `undefined`? Otherwise, you can omit the fallback here.

frontend/vre/utils/jsonld.model.js
```diff
@@ -1,11 +1,91 @@
...
+        const orderedItems = orderedCollection[`${this.activityStreamsPrefix}orderedItems`]["@list"]
```
jgonggrijp, line 83: Anywhere between lines 80 and 84, create your lookup structure like this:

```js
const subjectsByID = _.indexBy(allSubjects, '@id');
```

it will likely include a spurious `undefined` entry, but that is unproblematic.

frontend/vre/utils/jsonld.model.js
```diff
@@ -1,11 +1,91 @@
...
+        const result = orderedItems.map((subject) => {
+            const id = subject["@id"];
+            const orderedSubject = allSubjects.find((subject) => subject["@id"] === id);
+            return nestSubject(allSubjects, orderedSubject);
+        }, this);
```
jgonggrijp, line 88:
```suggestion
        const result = orderedItems.map((subject) => {
            const orderedSubject = subjectsByID[subject["@id"]];
            return nestSubject(subjectsByID, orderedSubject);
        });
```

I removed the `this` because it isn't used. Don't forget to update `nestSubject` to use the index instead of the graph!

frontend/vre/utils/jsonld.model.test.js
```diff
@@ -0,0 +1,81 @@
...
+    "example:value": {
+        "@id": "http://example.com/descForDescForS4"
+    }
+}, {
+    "@id": "http://example.com/descForDescForS4",
+    "example:value": "Random description",
+}, {
+    "@id": "http://example.com/descForS2",
+    "example:value": "Random description",
+}];
```
jgonggrijp, line 53:
```suggestion
}, {
    "@id": "http://example.com/s7",
    "dc:title": "Title with undetected cycle",
    "dc:description": {
        "@id": "http://example.com/descForS7",
    },
}, {
    "@id": "http://example.com/descForS7",
    "owl:sameAs": {
        "@id": "http://example.com/sameDescForS7",
    },
}, {
    "@id": "http://example.com/sameDescForS7",
    "owl:sameAs": {
        "@id": "http://example.com/descForS7",
    },
}];
```

frontend/vre/utils/jsonld.model.test.js
```diff
@@ -0,0 +1,81 @@
...
+        const subject = findById(exampleJsonLDGraph, "http://example.com/s4");
+        const ns = nestSubject(exampleJsonLDGraph, subject);
+        assert.equal("Random description", ns["dc:description"]["example:value"]["example:value"]);
+    });
+    it('does not resolve a recursive reference', () => {
+        const subject = findById(exampleJsonLDGraph, "http://example.com/s5");
+        const ns = nestSubject(exampleJsonLDGraph, subject);
+        assert.equal("http://example.com/s5", ns["owl:sameAs"]["owl:sameAs"]["@id"]);
+    });
+});
```
jgonggrijp, line 81:
```suggestion
    it('detects cycles even if the root subject is not involved', () => {
        const subject = findById(exampleJsonLDGraph, "http://example.com/s7");
        const ns = nestSubject(exampleJsonLDGraph, subject);
        assert.equal("http://example.com/descForS7", ns["dc:description"]["owl:sameAs"]["owl:sameAs"]["@id"]);
    });
});
```

frontend/vre/utils/jsonld.model.test.js
```diff
@@ -0,0 +1,81 @@
...
+}, {
+    "@id": "http://example.com/descForS2",
+    "example:value": "Random description",
+}];
+
+describe('nestSubject', () => {
+    it('does not change anything if there are no references', () => {
+        const subject = findById(exampleJsonLDGraph, "http://example.com/s1");
+        const ns = nestSubject(exampleJsonLDGraph, subject);
+        assert.equal(ns["dc:title"], "Title without references");
```
jgonggrijp, line 59: Did you know that if you write this,

```suggestion
        assert(ns["dc:title"] === "Title without references");
```

and the assertion fails, power-assert will tease apart the expression and show you all intermediate values, just like in pytest assertions?

There is nothing wrong with `assert.equal` of course, so you don't need to go change all your assertions, but it might still be nice to know that you can do the pytest thing in the frontend, too.

jgonggrijp (changes requested):
This looks good to me overall. I made a bunch of comments, but they are about details and not all of them require changes.

(Side note: the comments probably read best on the conversation tab, because the order in which I submitted the comments is not the same as the order in which they appear in the code.)

frontend/vre/record/record.list.item.view.js
```diff
@@ -1,28 +1,33 @@
...
+/*
+ * Note: the display of existing annotations has temporarily been
+ * removed from this view. Refer to git history to see how it was implemented.
+ */
+
 export var RecordListItemView = SelectableView.extend({
+    /**
+     * @type Record
+     */
+    model: null,
```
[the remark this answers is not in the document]
tijmenbaarda: I manually added the type hints to document what type is expected in the model. I don't know why I imported this.

By the way, I was assuming that unused imports would not end up in the bundle, so it is good to know that I should be wary of that.

frontend/vre/utils/jsonld.model.js
```diff
@@ -1,11 +1,91 @@
...
+ * @param subject{JSONLDSubject} - The subject including its predicates and objects to create a nested version of
+ * @returns {Object}
+ */
+export function nestSubject(graph, subject) {
+    const recursiveFunction = function(graph, subject, baseSubject) {
+        const transformedSubject = _.clone(subject);
+        for (let property of Object.keys(subject)) {
+            if (subject[property].hasOwnProperty("@id")) {
+                // This is a reference to another subject
+                const refereedSubject = graph.find((thisSubject) => thisSubject["@id"] === subject[property]["@id"]);
```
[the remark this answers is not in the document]
tijmenbaarda: Great! The `indexBy` function is really convenient, but unfortunately it took me quite some time to realize that it was renamed to `keyBy` in lodash

frontend/vre/record/record.list.item.view.js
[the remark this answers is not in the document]
jgonggrijp: I think you can pass `treeshake: true` to the Rollup config, but it isn't perfect. Better to not import things you don't use.

frontend/vre/utils/jsonld.model.js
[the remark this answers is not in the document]
jgonggrijp: Ah, should have double-checked before posting. Sorry for not doing that.

frontend/vre/utils/jsonld.model.js
tijmenbaarda: I did not think of this! I would not expect this to happen right now, but it certainly may at some point. Thank you for the additional test - it passes now.

frontend/vre/utils/jsonld.model.js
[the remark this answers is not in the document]
tijmenbaarda: No, I should have been more suspicious :)

frontend/vre/record/record.list.item.view.js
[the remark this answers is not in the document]
tijmenbaarda: OK!

frontend/vre/utils/jsonld.model.js
```diff
@@ -1,11 +1,91 @@
...
+
+/**
+ * Generic subclass of APICollection that parses incoming JSON-LD to an
+ * array of all subjects. The contents of subjects are left unchanged.
  */
 export var JsonLdCollection = APICollection.extend({
     parse: function(response) {
+        if (!response.hasOwnProperty("@graph")) {
+            throw "Response has no @graph key";
+        }
```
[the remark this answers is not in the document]
tijmenbaarda: Not now, because the backend always adds a context and returns the compacted form. I have added some notes that this compacted form is expected. However, I think that some additional robustness is needed, because I have only tested the function with data coming from rdflib... I have created issue #184 for this.

frontend/vre/main.js
```diff
@@ -69,6 +67,8 @@ var VRERouter = Backbone.Router.extend({
         GlobalVariables.currentCatalog = GlobalVariables.catalogs.findWhere({
             identifier: id,
         });
+        GlobalVariables.searchView.source = GlobalVariables.currentCatalog.get("@id");
```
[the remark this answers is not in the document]
tijmenbaarda: Cool!

frontend/vre/utils/jsonld.model.test.js
tijmenbaarda: Great, I didn't know that!

backend/catalogs/graphs.py
```diff
@@ -105,12 +105,12 @@ def set_query(
...
             query: str,
             start: int = 0,
-            end: int = 50
+            end: Optional[int] = None
     ):
         """Set the query. Should be called before calling
         ``perform_fetch()``."""
         self._start = start
-        self._end = end
+        self._end = end if end is not None else start + 50
```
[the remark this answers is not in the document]
tijmenbaarda: Changed this

tijmenbaarda:
Thank you for the thorough review, I learned a lot from it! I have made all the requested changes, I think.

frontend/vre/utils/jsonld.model.js
```diff
@@ -1,11 +1,102 @@
...
+
+/**
+ * An individual subject definition inside JSON-LD
+ * @typedef {Object} JSONLDSubject
+ */
+
+export var JsonLdModel = Backbone.Model.extend({
+    idAttribute: '@id',
+
+})
```
jgonggrijp, line 16:
```suggestion
});
```

frontend/vre/utils/jsonld.model.js
```diff
@@ -1,11 +1,102 @@
...
+ * @param parentSubjectIDs{Array<String>} - For internal use of recursive function; leave undefined
+ * @returns {Object}
+ */
+export function nestSubject(subjectsByID, subject, parentSubjectIDs=undefined) {
+    if (typeof parentSubjectIDs === "undefined") {
+        parentSubjectIDs = [subject["@id"]];
+    } else {
+        parentSubjectIDs = Array.from(parentSubjectIDs);
+        parentSubjectIDs.push(subject["@id"]);
+    }
```
jgonggrijp, line 47:
```suggestion
export function nestSubject(subjectsByID, subject, parentSubjectIDs=[]) {
    parentSubjectIDs.push(subject["@id"]);
```

frontend/vre/utils/jsonld.model.js
```diff
@@ -1,11 +1,102 @@
...
+    const transformedSubject = _.clone(subject);
+    for (let property of Object.keys(subject)) {
+        if (subject[property].hasOwnProperty("@id")) {
+            // This is a reference to another subject
+            const refereedSubject = subjectsByID[subject[property]["@id"]];
+            if (refereedSubject && !(parentSubjectIDs.includes(refereedSubject["@id"]))) {
+                /* If the refereed subject was found in the graph, use it as replacement.
+                   Only do this if the subject is not the same as the one we started with,
+                   to avoid an endless loop. (Alternative would be to create a circular reference) */
```
jgonggrijp, line 56:
```suggestion
                   Only do this if we have not visited the same subject before,
                   to avoid an endless loop. (Alternative would be to create a circular reference) */
```

frontend/vre/utils/jsonld.model.js
```diff
@@ -1,11 +1,102 @@
...
+     * @type {string}
+     */
+    activityStreamsPrefix: "as:",
+    parse: function(response) {
+        // Get all subjects of the graph with their predicates and objects as an array
+        if (!response.hasOwnProperty("@graph")) {
+            throw "Response has no @graph key, is this JSON-LD in compacted form?";
+        }
+        const allSubjects = response["@graph"];
+        const orderedCollection = allSubjects.find((subject) => {return subject["@type"] === `${this.activityStreamsPrefix}OrderedCollection`}, this);
```
jgonggrijp, line 92: I think I forgot to mention this in my previous review, but Underscore's `_.find` lets you write  a nice shorthand:

```suggestion
        const orderedCollection = _.find(allSubjects, {"@type": `${this.activityStreamsPrefix}OrderedCollection`});
```

With an additional variable, the code is even easier to read:

```suggestion
        const ocType = `${this.activityStreamsPrefix}OrderedCollection`;
        const orderedCollection = _.find(allSubjects, {"@type": ocType});
```

jgonggrijp (commented):
Looks good to me. Just a few nitpicks and one more tip.

frontend/vre/utils/jsonld.model.js
tijmenbaarda: Hmm, I wrote it this way to make sure the array is copied so that it is not altered by upcoming invocations of the function. Is that not a problem in JavaScript?

jgonggrijp: Two things:

1. In Python, you must never default a function parameter to something mutable like a list, because that default value is created at function definition time rather than function invocation time. That problem does not exist in most other programming languages. In JavaScript, you can safely do `someParameter = []`.
2. You are right that the same array is reused when you don't create a new one by default. This means that you also have to do `parentSubjectIDs.pop()` at the end of the function. Sorry, did not pay attention to whether you were already doing that.

tijmenbaarda: Ah, thank you! I have applied your changes and added the `pop` call at the end of the function.

tijmenbaarda:
Everything has been addressed now, I will merge the pull request.

PR #192, opened 2024-07-22
Build the interface on sync, not on update

### LINKED ISSUE #189: Part of the interface is not shown if there are no collections
If there are no collections at all, the catalog and collections dropdowns do not show. No exception is raised.

Desired behaviour: the dropdown should be empty with a hint that there are currently no collections. (Creating collections is not possible from within the UI for now.)

After fixing this bug, also verify that the interface works correctly if there are no projects.

tijmenbaarda:
The function `startRouting` was never called if there were zero collections or projects, because the backend would return empty arrays and the `update` event was never fired.

Closes #189. @jgonggrijp : can you verify that changing `update` to `sync` indeed has the desired effect?


jgonggrijp (approved):
Yes, I think you are right that this change should sort the desired effect. The `'sync'` event happens after the `'update'` event, so by the time those three `'sync'` events have triggered, the router should have at least as much information to go by as after the three corresponding `'update'` events.

Reading over this code again, I realize that lines 116-120, where the dropdowns and the empty record button are placed, could also be moved up into the `prepareCollections` function. Those views are self-updating, so their placement only depends on the DOM being ready and not on the collections being complete.

This change would also solve the problem that those views do not appear when there is an empty set of collections. However, your replacement of `'update'` by `'sync'` would still be necessary in order to ensure that the router will do its job.

Clever solution, well done!

PR #197, opened 2024-07-24
Feature/fixtests

tijmenbaarda:
A few unit tests were not working anymore.

@jgonggrijp : it appeared that the function `remove_from_triplestore` was not working with subjects that had no references to other subjects in the graph. I tried to fix it, and the unit tests are working now (I also added a new test with a subject that *did* refer to another subject, which was already passing), but I don't know if the change I made in the sparql query is indeed correct. Could you take a look at that?


backend/catalogs/triplestore.py
```diff
@@ -22,7 +22,7 @@
   values ?r {{ {obsolete_records} }}
   ?r ?p1 ?o1;
      ?pt ?f.
-  ?f ?p2 ?o2.
+  optional {{?f ?p2 ?o2.}}
```
jgonggrijp, line 25: This change is correct, *if* we consider it acceptable for a record to have no fields. Otherwise, this would amount to working around a data integrity issue.

backend/catalogs/triplestore_test.py
```diff
@@ -1,32 +1,38 @@
...
+    return [record0, record1]
+
+
+@pytest.fixture
+def working_data_graph(working_data_records) -> tuple[list[Record], Graph]:
+    record0, record1 = working_data_records
+    """Create some data to work with, as a list of records (needed for
+    removal) and as a graph."""
     graph = record0.to_graph() + record1.to_graph()
     return [record0, record1], graph
```
jgonggrijp, line 24: I like how you split this fixture into two smaller, more reusable pieces.

jgonggrijp (approved):

backend/catalogs/triplestore.py
tijmenbaarda: Ah, I see. A record with no fields is not very useful, but it is not incorrect either. You could think of a record for which no normalization at all has been defined, but which still has a URI, the ID of the original database and the original data.

PR #199, opened 2024-07-31
Show EDPOP Explorer record details

tijmenbaarda:
The display names of the record fields are taken from the EDPOP record ontology.

I created a simple reader for these property names -- I realized that to do it properly will get quite complex very soon. My main doubt here is where to put the ontology (or, a JSON-LD version of it):
* Compile it into the JavaScript bundle (probably the most efficient)
* Take it live from dhstatic
* Add it as a static file and read it from there (my current approach - this was the easiest to implement)

### DIFF
--- a/frontend/vre/css/style.css
+++ b/frontend/vre/css/style.css
@@ -35,3 +35,16 @@ table#hpb-info-table td:nth-child(odd) {
 		width: 636px;
 	}
 }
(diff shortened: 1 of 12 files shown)


lukavdplas (approved):
I didn't check the interface, but the code looks good!

PR #200, opened 2024-08-01
Tests using GitHub Actions

tijmenbaarda:
I added these because we have merged branches with failing tests a couple of times (mea culpa)


.github/workflows/test.yml
```diff
@@ -0,0 +1,77 @@
...
+      - 'main'
+      - 'release/**'
+      - 'feature/ci'
+  pull_request:
+    branches:
+      - 'develop'
+      - 'main'
+
+jobs:
+  build:
```
lukavdplas, line 17: "build" is an odd name for this job, "test" would be more suitable.

.github/workflows/test.yml
```diff
@@ -0,0 +1,77 @@
...
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: ["3.9", "3.10", "3.11", "3.12"]
+        node-version: ['20.x']
```
lukavdplas, line 22: If we're only testing one version of node, it doesn't need to be in a matrix.

.github/workflows/test.yml
```diff
@@ -0,0 +1,77 @@
...
+    branches:
+      - 'develop'
+      - 'main'
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: ["3.9", "3.10", "3.11", "3.12"]
```
lukavdplas, line 21: I'm not convinced it's really all that useful to test four different Python versions; you could just test the version that runs on the production server.

.github/workflows/test.yml
```diff
@@ -0,0 +1,77 @@
...
+          cache: 'npm'
+          cache-dependency-path: './frontend/package-lock.json'
+      - name: Install frontend dependencies
+        run: |
+          cd frontend
+          npm ci
+      - name: Run frontend tests
+        run: |
+          cd frontend
+          npm run test
```
lukavdplas, line 77: I think this is completely independent of the backend installation + tests in previous steps, so you could split these steps into a separate job (currently, everything is a single "build" job). That way, the frontend and backend tests can be run in parallel, so the action will run faster.

lukavdplas (approved):
Nice!

I've added some suggestions, but none of them are required.

.github/workflows/test.yml
tijmenbaarda: Makes sense :)

.github/workflows/test.yml
tijmenbaarda: Four versions may indeed be overkill, but I'll keep 3.12 to facilitate installation at other places.

.github/workflows/test.yml
tijmenbaarda: OK!

frontend/karma.conf.js
```diff
@@ -93,14 +93,32 @@ module.exports = function(config) {
 
     // start these browsers
     // available browser launchers: https://www.npmjs.com/search?q=keywords:karma-launcher
-    browsers: [],
+    browsers: ['ChromeHeadless'],
```
jgonggrijp, line 96: If I'm right, this change causes Karma to attempt to launch headless Chrome when we run the tests locally in development, too. That would be inconvenient for me, firstly because I have so far managed to avoid installing Chrome and secondly because I like to debug the tests in a headed browser.

I might eventually submit a PR to undo the change on this line, because line 119 already ensures that headless Chrome is used in CI.

tijmenbaarda: Oh, I am sorry for that, that was not my intention. I accidentally kept that line, which I added when I was debugging this. (By the way, `ChromeHeadless` works also using Chromium, but I don't know if that is your concern about installing Chrome.)

PR #214, opened 2024-08-20
Use Tabulator library for list view

tijmenbaarda:
I have been experimenting using the Tabulator JavaScript library for the record list view. This seems to work very well, and it allows for a multi-column list view, which is one of the new features we promised. It also supports sorting out of the box.

Things that should still be added:
* Filtering (this can be customized, so that we can for example add a year range filter that works on normalized years) - this removes the need for customized query options for a great deal, except perhaps in the case where the number of search results is extremely large
* Showing and hiding of columns
* Export to CSV/XSLX/PDF

Tabulator seems very well maintained and documented and it has an MIT license, but it is new for me and I don't know about alternatives. For this review, apart from taking a look at the code, could you let me know if there are any doubts about using Tabulator for the list view?

I have imported Tabulator by adding it to NPM, but this does not work for the CSS part because our Rollup configuration does not bundle CSS. @jgonggrijp : do you know what the best solution is here? As an alternative, we might also use the CDN version.


frontend/vre/index.html
```diff
@@ -12,6 +12,7 @@
         >
         <link rel="stylesheet" type="text/css" href="/static/css/style.css">
         <link href="https://cdnjs.cloudflare.com/ajax/libs/select2/4.0.13/css/select2.min.css" rel="stylesheet" />
+        <link href="https://unpkg.com/tabulator-tables@6.2.5/dist/css/tabulator.min.css" rel="stylesheet">
```
lukavdplas, line 15: Looks like Tabulator also has a [bootstrap 3 theme](https://tabulator.info/docs/6.2/theme#framework-boot3); that will probably look better in the interface.

lukavdplas (approved):
Neat! Using Tabulator sounds like a good choice to me.

frontend/vre/index.html
tijmenbaarda: Cool, it looks much better with it indeed!

jgonggrijp:
@tijmenbaarda When would you prefer to receive my review at the latest?

tijmenbaarda:
> @tijmenbaarda When would you prefer to receive my review at the latest?

I would like to build further on this as soon as possible... If a review is not possible now and you don't see worrying things at first sight, I would suggest we merge this now, as Luka has already looked at it. In any case, thanks for the quick reply!

frontend/vre/record/record.list.view.js
```diff
@@ -1,61 +1,68 @@
...
+import {getStringLiteral} from "../utils/jsonld.model";
+import {vreChannel} from "../radio";
 
-export var RecordListView = CollectionView.extend({
-    tagName: 'table',
-    className: 'table table-striped',
-    template: recordListTemplate,
-    container: 'tbody',
+export var RecordListView = Backbone.View.extend({
+    tagName: "div",
```
jgonggrijp, line 8: For your information: `'div'` is the default, so you don't strictly need to set the `tagName` when you choose this value. It does not hurt to set it, though.

backend/boek.json
jgonggrijp, line 1: Is this file supposed to be part of the PR? What is it doing?

jgonggrijp (approved):
> Tabulator seems very well maintained and documented and it has an MIT license, but it is new for me and I don't know about alternatives. For this review, apart from taking a look at the code, could you let me know if there are any doubts about using Tabulator for the list view?

I don't know it either, but I share your impression that it is well maintained and batteries included. The only thing that worries me slightly is that you update it by replacing the data wholesale. However, given how much code you're able to remove, I am still in favor of the change.

> 
> I have imported Tabulator by adding it to NPM, but this does not work for the CSS part because our Rollup configuration does not bundle CSS. @jgonggrijp : do you know what the best solution is here? As an alternative, we might also use the CDN version.

We are not using Sass in this project, are we? If that were the case, it would just be a matter of an `@import`. Without a proper CSS build tool, I think CDN is a fine way to go.

backend/boek.json
jgonggrijp: Generally when a file is accidentally committed, it is better to rebase the branch so that the file completely disappears from history. For this time you can leave it at this, though.

frontend/vre/record/record.list.view.js
```diff
@@ -1,5 +1,4 @@
 import Backbone from "backbone";
-import {TabulatorFull as Tabulator} from "tabulator-tables";
```
jgonggrijp: Strictly speaking, you still need this import. You can make Rollup do the right thing by adding `tabulator-tables` to the [external](https://rollupjs.org/configuration-options/#external) and the [output.globals](https://rollupjs.org/configuration-options/#output-globals) in the `rollup.config.js`.

PR #223, opened 2024-10-02
Add basic filters to search view

tijmenbaarda:


jgonggrijp (approved):
I scrolled through the code and tried the tables that resulted from searches in the KB and the HPB. It seems to work well.

Small comment: the title filter input is larger than the filter inputs of the other columns. This makes intuitive sense, but it looks a little bit disorganized, also because the border of the input (textarea?) looks slightly different. I suggest making this input of the same type as the other columns, but this is not a requirement for approval.

Larger comment: pattern searching would be convenient, perhaps regex-based. That could certainly wait, though.

Overall, I like it a lot!

Issue #226, opened 2024-10-03
Migrate old-style collections to triplestore


jgonggrijp:
I merged #181, started working on #224 and then realized that all my collections were gone. Or rather, they don't appear in the frontend anymore because my old testing collections are stored in the relational database while the backend now attempts to retrieve them from the triplestore.

This needs to be addressed before we can deploy from the `develop` branch.

lukavdplas:
There is a [convert_legacy_data](https://github.com/UUDigitalHumanitieslab/EDPOP/tree/develop/backend/triplestore/convert_legacy_data) module in the backend which defines functions for this, but it's not used.

I believe the reason why collections are not migrated in #181 was the interdependencies between records, collections, and annotations; this makes it difficult to migrate one model but not the other.

jgonggrijp:
Makes sense. That does mean we have to implement migrations for all three before we can deploy off `develop` again.

jgonggrijp:
@tijmenbaarda I think we can close this as of #311 and #324?

tijmenbaarda:
I think so too!

Issue #225, opened 2024-10-03
Manage collections from the frontend


jgonggrijp:
Unlock at least some of the features of the new collections endpoint (#181) from the frontend. I will be looking into this together with #224.

PR #230, opened 2024-10-09
Basic frontend collection management

### LINKED ISSUE #224: Rewire frontend with backend collections endpoint
After #181, this needs to happen.

jgonggrijp:
While frontend collection handling forms the overarching story of this branch, it is a bit of a hodgepodge of small changes:

- Fixed #224 (restores ability to list and display collections).
    - Adjusted existing models, views and other parts to the changed field names.
    - Changed the project menu to automatically switch to the project of the current collection, since each collection can now be associated with only one project.
    - Made the `summary` of a collection visible through a `<small>` insertion in the page header.
    - Abolished the conceptual difference between "my collections" and "all collections", because the backend endpoint will never list collections beyond the ones the user has access to.
    - Set the `idAttribute` of the Backbone model that represents a single VRE collection to `uri`.
- Minimally implemented #225 (create and partial update only).
    - Added an inline form to the bottom of the collection menu that lets the user quickly whip up new empty collections.
    - Added an inline form that can be temporarily overlaid over the `<small>` in the page header in order to adjust the `summary` of a collection.
        - Created a separate, generic, reusable `OverlayView` in order to support this. It can do one thing: temporarily replace an arbitrary DOM element by an arbitrary view, based on a selector relative to a persistent root element.
- Other fixes, refactorings, cosmetic changes and enhancements (mostly just to give myself an easier time while working on the above changes).
    - Enabled [alt-click debugging](https://github.com/CentreForDigitalHumanities/backbone-util/blob/develop/doc/click-to-debug.md) for all views (with the exception of views that aren't reachable by mouse, such as `OverlayView` instances).
    - Set the `idAttribute` of the `Project` model to `name`. It appears that the `idAttribute` was never set to a correct value before this.
    - Added an `.editorconfig`.
    - Added temporary files generated by emacs (which I use) to the `.gitignore`.
    - Renamed the "annotation context" menu to just "project", because I felt this conveys the meaning of the menu better.
    - Wrote "Collection" instead of "Browse or search" in the header of the collection page, because I think this signals more clearly to the user what they are viewing.
    - Added a provisional auto dark mode.
    - Removed `headlessChrome` from the default browsers in the Karma config as discussed in https://github.com/UUDigitalHumanitieslab/EDPOP/pull/200#discussion_r1740803381.
    - Removed the project menu and associated collections from the `GlobalVariables`, using the `vreChannel` instead for loose coupling. This brings us a bit closer to #123 again.
    - Replaced the old, simple but faulty logic for appending trailing slashes to model URLs by the more sophisticated and thoroughly tested [`modelSlashUrl`](https://github.com/CentreForDigitalHumanities/backbone-util/blob/develop/doc/trailing-slash.md) from backbone-util.

I will be on vacation soon, so I prioritized fast delivery over comprehensiveness. Hence, there are some omissions and rough edges that I am aware of:

- Collections cannot be deleted. I think this should eventually be possible, but it needs to be designed carefully to minimize the risk that users accidentally lose a lot of work.
- The name of a collection cannot be changed. I opted out of this because I did not want the name and the uri to diverge. In the future, users could conceivably work around this by adding all records from one collection to another and then deleting the source collection.
- A project must be selected before you can create a collection. If no project is selected, collection creation silently fails, ~~but an entry with the new name still appears in the menu.~~
- There is no validation or error handing for the create and update actions.

There are also things that were already broken on `develop` and which are not fixed by this branch:

- Collections created before #181 are not visible because of #226.
- There is no way to add or list records in collections because the records endpoint still has to be updated.
- Project selection should probably be more prominent.

For review, I suggest giving the frontend a quick spin and just scrolling through the code changes. There are lots of files with only minor changes, but I think you will be able to quickly identify repeating patterns and skip over those.

### DIFF
--- a/frontend/vre/project/project.menu.view.js
+++ b/frontend/vre/project/project.menu.view.js
@@ -50,6 +50,6 @@ export var ProjectMenuView = AggregateView.extend({
         this.model = model;
         this.renderContainer();
         this.trigger('select', model);
-        localStorage.setItem('project', model.attributes.id);
+        localStorage.setItem('project', model.id);
     },
 });
--- a/frontend/vre/project/project.menu.view.js
+++ b/frontend/vre/project/project.menu.view.js
@@ -40,6 +40,9 @@ export var ProjectMenuView = AggregateView.extend({
     },
 
     select: function (model) {
+        // At this point, `model` could either be an instance of `Project` or
+        // just an id. The next line ensures that it is a full-blown instance
+        // (or `undefined`).
         model = this.collection.get(model);
         if (!model) return;
         if (model === this.model) return;
--- a/frontend/vre/project/project.menu.view.js
+++ b/frontend/vre/project/project.menu.view.js
@@ -40,6 +40,8 @@ export var ProjectMenuView = AggregateView.extend({
     },
 
     select: function (model) {
+        model = this.collection.get(model);
+        if (!model) return;
         if (model === this.model) return;
         if (this.model) this.model.trigger('deselect');
         this.model = model;
--- a/frontend/vre/project/project.menu.view.js
+++ b/frontend/vre/project/project.menu.view.js
@@ -1,9 +1,9 @@
-import { CollectionView } from 'backbone-fractal';
+import { AggregateView } from '../core/view.js';
 
 import { ProjectMenuItemView } from './project.menu.item.view';
 import projectMenuTemplate from './project.menu.view.mustache';
 
-export var ProjectMenuView = CollectionView.extend({
+export var ProjectMenuView = AggregateView.extend({
     el: '#vre-project-menu',
     template: projectMenuTemplate,
     subview: ProjectMenuItemView,
--- a/frontend/vre/utils/overlay.view.js
+++ b/frontend/vre/utils/overlay.view.js
@@ -1,0 +1,45 @@
+import _ from 'lodash';
+import { View, $ } from 'backbone';
+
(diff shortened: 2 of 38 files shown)


frontend/vre/utils/overlay.view.js
```diff
@@ -0,0 +1,45 @@
...
+    uncover: function() {
+        if (!this.covered) return false;
+        if (!this.root.find(this.guest.el).length) throw Error(guestAbsent);
+        this.guest.$el.before(this.covered).detach();
+        delete this.covered;
+        return true;
+    },
+
+    toggle: function() {
+        this.cover() || this.uncover();
```
tijmenbaarda, line 32: Wow!

frontend/vre/project/project.menu.view.js
```diff
@@ -40,11 +40,13 @@ export var ProjectMenuView = CollectionView.extend({
     },
 
     select: function (model) {
+        model = this.collection.get(model);
```
tijmenbaarda, line 43: Not very clear, I would change the variable name or add a comment

tijmenbaarda:
> * A project must be selected before you can create a collection. If no project is selected, collection creation silently fails, but an entry with the new name still appears in the menu.

When I try this, no new entry is added to the menu but an exception `project is undefined` is raised.

tijmenbaarda (approved):
Wonderful! I love the `OverlayView`. I have just a few comments. I can pick up the things of which you mention that they are still missing in the time that you are away.

Issue #229, opened 2024-10-04
Always show the same set of fields per record type


tijmenbaarda:

jgonggrijp:
I think you mean when displaying a record in the frontend?

tijmenbaarda:
Yes! Both in the table view and the list view a fixed set of fields should be shown. This should probably be the fields as defined in the record ontology. Currently the table view shows all fields that are present in one or more records, and the detail view only shows the fields that are present in that particular record.

In the table view some fields may be hidden by default, but they should always be defined.

This is per Jeroen's request, because he wants to be able to annotate fields that are not filled in.

jgonggrijp:
@tijmenbaarda in https://github.com/UUDigitalHumanitieslab/EDPOP/pull/233#pullrequestreview-2464458280:

> What I meant with #229 (and maybe I had not formulated it clearly) was also to show _only_ the fields that belong to the specific record type (like extent, which would only show on `BibliographicalRecord`). However, I think that the code in this PR prepares well for this with the functions in `tabulator-utils.js`.

Do you mean you also want to exclude the only-biographical fields from the hidden-but-available columns in the tabulator view of an only-bibliographical catalog (and vice versa)? Because in that case we need a way to distinguish between only-biographical, only-bibliographical and mixed catalogs. Likewise for collections.

Writing this out, I realize this probably isn't worth the effort. We can still do it for the record detail view, of course.

Issue #228, opened 2024-10-04
Make a clear distinction between biographical and bibliographical catalogues


tijmenbaarda:

jgonggrijp:
... in the frontend? With a little pill/tag or something?

PR #233, opened 2024-11-13
Make all properties available in Tabulator records list, show record type with icon

jgonggrijp:
This implements half of #229 and half of #228. Showing record types or all properties in the detail view is not yet implemented, although some of the logic introduced in this branch could be reused for that purpose. I wanted to go all the way, but did not manage to go beyond the record list because of time constraints and distractions.

The `autoColumnsDefinitions` callback from `frontend/vre/record/record.list.view.js` was outfactored to `frontend/vre/utils/tabulator-utils.js` and then extended to always include all possible properties as available columns. The set of *visible* columns is still determined in the same way, though the decision to make a column visible now happens in a different place. For the extended logic, I used the `MappedCollection` feature from the freshly outfactored, still-to-be-released [backbone-collection-transformers](https://github.com/CentreForDigitalHumanities/backbone-collection-transformers).

Then, I added an icon column to convey the record type: a person silhouette for a biographical record, a book for a bibliographical record. The HTML for the icon is rendered using a Mustache template. That template is potentially reusable for a `Backbone.View`. The HTML includes textual labels for screen readers. Record types are indicated at the individual record level rather than the catalog level, because I seemed to recall that some catalogs are mixed.

The new code lacks tests, but it has a low complexity and I hope it is sufficiently documented.

This branch has some side effects:

- The docker-compose `postgres` service port is no longer forwarded to the host. This was never necessary to make docker-compose work. The reason to remove it is that it somehow prevented me from starting the container.
- I turned the filter input for the Title column from a `<textarea>` back into a default `<input>`, because the height difference with other columns annoyed me.
- I fixed a typo in the `edpop-record-ontology.json` that I discovered by coincidence.

As an aside, I notice that `edpoprec:BiographicalRecord` is a subclass of `edpoprec:Record` while `edpoprec:BibliographicalRecord` isn't. Conversely, `BibliographicalRecord` has a cardinality constraint while `BiographicalRecord` doesn't. I did not address this because I didn't know whether it was merely an oversight or whether there was a good reason for it.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -11,8 +11,6 @@ services:
         volumes:
             - pgdata:/var/lib/postgresql/data
             - ./backend/create_db.sql:/docker-entrypoint-initdb.d/edpop.sql
-        ports:
-            - 127.0.0.1:5432:5432
     blazegraph:
         image: islandora/blazegraph:main
         healthcheck:
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -9,6 +9,7 @@
       "version": "1.0.0",
       "license": "MIT",
       "dependencies": {
+        "@uu-cdh/backbone-collection-transformers": "github:CentreForDigitalHumanities/backbone-collection-transformers#release/0.1.0",
         "@uu-cdh/backbone-util": "^0.1.1",
         "backbone": "^1.3.3",
         "backbone-fractal": "^1.1.1",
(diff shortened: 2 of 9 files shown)


tijmenbaarda (approved):
I have no comments on the code, but as always it was a pleasure to read and I learned a lot from it.

What I meant with #229 (and maybe I had not formulated it clearly) was also to show *only* the fields that belong to the specific record type (like extent, which would only show on `BibliographicalRecord`). However, I think that the code in this PR prepares well for this with the functions in `tabulator-utils.js`.

tijmenbaarda:
> As an aside, I notice that `edpoprec:BiographicalRecord` is a subclass of `edpoprec:Record` while `edpoprec:BibliographicalRecord` isn't. Conversely, `BibliographicalRecord` has a cardinality constraint while `BiographicalRecord` doesn't. I did not address this because I didn't know whether it was merely an oversight or whether there was a good reason for it.

This must have been an oversight of mine. I am busy with an updated record ontology and I will take this into account, too.

Issue #236, opened 2024-12-04
Annotations (birds-eye view)


jgonggrijp:
Based on the email exchange between @tijmenbaarda and Jeroen Salman on November 18-19:

- Annotations may take three forms.
    1. Correcting a field's value.
    2. Adding a missing field.
    3. General comment to the record as a whole.
- Annotations are specific to the project but not to the collection. All collections within a project see the same annotations.
- It must be possible for two different users to make separate (potentially even conflicting) annotations on the same field of the same record within the same project.
- When a field has multiple values (or can have multiple values), it must be possible to annotate (i.e., correct or add) each value separately.

Regarding the latter point, it is probably helpful to think not of a field with multiple values, but of multiple field-value pairs that happen to have the same field type.

tijmenbaarda:
In addition, in our meeting on 20 December, Jeroen noted that it should be possible to add

Issue #232, opened 2024-11-06
Correctly handle repeated fields


tijmenbaarda:
Both in the table view and in the detail view, fields that are repeated (such as contributor, if there are multiple contributors) are currently not shown at all.

In the table view they can probably just be shown with a separator in between.

In the detail view something more elaborated may be needed (because we want to be able to annotate them separately), but for now a separator (or line break?) may be just fine.

PR #242, opened 2024-12-12
Repeated fields

### LINKED ISSUE #240: Frontend Record instances with multiple contributors are incomplete
I was orienting myself on #232 and needed an example of a record with a repeated field. With help from @tijmenbaarda, I eventually found a record in the USTC with the title *The Bucoliks otherwise called his pastoralls, or shepeherds meetings. Together with his Georgiks or ruralls, otherwise called his husbandrie, conteyning foure books*, which has three contributors. It can be uniquely found by searching for "Bucoliks".

I was using the alt-click debugging feature in the frontend to inspect the contents of the record model. As expected, the `edpoprec:contributor` attribute had a three-element array as value, but there was an unpleasant surprise. The three objects in the array are blank nodes with *only* an `@id`; the `@type` and the `edpoprec:originalText` are absent, even though those appear to be always present in singleton contributors.

Schermafdruk_20241211_191723

This means that additional information would have to be retrieved from elsewhere in order to actually display the names of the contributors. I have not investigated whether the same problem occurs with other repeated fields, but this seems likely.

Since this effectively blocks #232, I will be investigating this myself.

jgonggrijp:
Fixes #240 and implements #232. I took #236 into account: separate field values can be individually clicked for annotation, although the actual annotating has not been implemented yet.

It may seem as if rewrote the `nestSubject` function from scratch. I did not; the change was spread over two commits where each commit was an evolutionary change from the previous situation.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,5 +1,3 @@
-version: "3.1"
-
 services:
     postgres:
         image: postgres:14
--- a/frontend/vre/field/field.model.js
+++ b/frontend/vre/field/field.model.js
@@ -42,6 +42,18 @@ export var Field = Backbone.Model.extend({
 });
 
 /**
+ * Find the set of properties that apply to the given record.
+ * Mostly an implementation detail of {@link FlatFields} and
+ * {@link FlatterFields}.
+ */
+function selectProperties(record) {
+    return (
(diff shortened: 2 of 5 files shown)


tijmenbaarda (approved):
Looks beautiful!

PR #247, opened 2025-02-10
Some catalogue view improvements

tijmenbaarda:
A few improvements to the catalogue view based on a conversation with Jeroen:

* Better displayment of errors that occur in the explorer: the messages of `ReaderError` are shown to the user
* Do not show tables when there are no results. This also shows the problem that when no results were shown, next queries were resulting in a table with rows but without columns
* Move the create blank record button, which was always visible but not always relevant, to the record list manager view (as per Jeroen's request)
* A small refactoring: rename `CollectionSearchView` to `CatalogSearchView`

### DIFF
--- a/backend/catalogs/api.py
+++ b/backend/catalogs/api.py
@@ -1,3 +1,4 @@
+from edpop_explorer import ReaderError
 from typing import Optional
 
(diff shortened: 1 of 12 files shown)


jgonggrijp (approved):
Looks sensible and conventional to me. :+1:

PR #250, opened 2025-02-11
Only show Field properties in table

tijmenbaarda:
At the moment, when selecting columns by right-clicking the column header, all properties of the ontology are shown, including `edpoprec:originalText` and other irrelevant properties. Only show properties that directly belong to any of the two record types.

### DIFF
--- a/frontend/vre/utils/record-ontology.js
+++ b/frontend/vre/utils/record-ontology.js
@@ -31,6 +31,7 @@ var recordTypePattern = /^edpoprec:((Bi(?:bli)?o)graphical)?Record$/;
  * @param {string} target - Can be `'Biblio'` or `'Bio'`, to indicate which
  * record type is considered acceptable. Any other value effectively means that
  * only the more generic type `'edpoprec:Record'` is acceptable.
+ * If not given, both types as well as the generic type are accepted.
  * @param {object} domain - JSON-LD declaration of a single domain of a
  * property.
  * @returns {boolean} `true` if the domain matches `target` or if it is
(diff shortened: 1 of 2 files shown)


frontend/vre/utils/record-ontology.js
```diff
@@ -31,6 +31,7 @@ var recordTypePattern = /^edpoprec:((Bi(?:bli)?o)graphical)?Record$/;
  * @param {string} target - Can be `'Biblio'` or `'Bio'`, to indicate which
  * record type is considered acceptable. Any other value effectively means that
  * only the more generic type `'edpoprec:Record'` is acceptable.
+ * If not given, both types as well as the generic type are accepted.
```
jgonggrijp, line 34: That means the previous sentence should be removed. :-)

frontend/vre/utils/record-ontology.js
```diff
@@ -39,9 +40,13 @@ var recordTypePattern = /^edpoprec:((Bi(?:bli)?o)graphical)?Record$/;
...
-    var qualified = match[1];
-    var qualifier = match[2];
-    return (!qualified || qualifier === target);
+    if (target) {
+        var qualified = match[1];
+        var qualifier = match[2];
+        return (!qualified || qualifier === target);
+    } else {
+        return true;
+    }
```
jgonggrijp, line 49: Clever modification, very observant. :+1:

jgonggrijp (changes requested):
LGTM, just one small doc thingy.

Issue #142, opened 2023-05-11
Manage collections as WIPs


lukavdplas:
Users will probably not create collections perfectly in a single session. Decent UI may be based on the following endpoints:

- Make a blank collection
- Import data from catalogi into a collection (#141 )
- Remove records from a collection individually (#17 
- Remove records based on a query
- Clone a collection

jgonggrijp:
All implemented except for removing records based on a query. They can be removed in bulk based on a filtering criterion in the table view.

Issue #29, opened 2018-03-19
Delete annotations


BeritJanssen:
Might display a bin icon next to an annotation.

PR #260, opened 2025-03-20
Once again, we can add records to collections and see the contents

jgonggrijp:
This implements most of #141 and #142, except for the ability to add all search results of a catalog query to a collection at once. Also not yet implemented are #29 (though removal is partly supported in the backend) and #226. There is a function that can do garbage collection, with tests, but no management command yet to kick off the garbage collection from the outside.

There is a lot of code churn. In many cases, I replaced old code that no longer worked (or was no longer sufficient) by similar code that does work (or is sufficient). The overall amount of code grew mostly because of new tests.

Some things that deserve clarification or special attention:

- Garbage collection uses a dedicated `records-gc` graph and takes a two-legged approach.
    - When a record (re)enters the triplestore because of a catalog search, it gets an associated `schema:lastUpdated` date stamp. I opted out of detailed timestamps because day resolution is sufficient for this purpose and it seemed more privacy-friendly. Only the date of the most recent fetch is kept.
    - As records are added to and removed from collections, a `schema:upvoteCount` keeps track of the number of collections the record is a member of. (This term choice is admittedly a bit cheesy; I couldn't find anything more fitting and didn't want to invent a new property for the purpose. I'm open to other options.) The parts of code where this tracking happens are completely disjoint from the pieces of code where the most recent fetch is tracked.
    - Only the actual garbage collection function takes both bits of information into account. It only removes records that are not in any collection and that were retrieved before a given cutoff date. That date defaults to two weeks before the date of invocation.
- `backend/collect/rdf_models.py` has three complex SPARQL queries (partly because of garbage collection) that strongly overlap. I outfactored the common parts to separate strings. Due to two layers of formatting, single braces in the final SPARQL appear as quadruple braces in some places. There are also double braces in some places. It's honestly quite messy and I'm open to alternatives.
- I generalized the frontend `nestSubject` function, which would collect nested resources for a single subject, to the `enforest` function, which does the same job for a whole array of subjects at once. This has a few advantages:
    - It simplifies the usage of the function, because the client code no longer needs to generate an index or iterate over the raw data.
    - It is more efficient, because every object in the array needs to be traversed only once. Cycle detection is easier as well. I did not actually find any reason to believe that the pre-existing code was too slow during testing, but it *might* make a difference if users manage to build collections with thousands of records.
    - It can also traverse `@list`s.
    - It makes it possible to autodetect which subjects are the outermost. More on this below.
- The frontend `JsonLdNestedCollection` class used to rely on a fixed `targetClass` in order to parse a flattened JSON-LD graph. Using `enforest` and a few tweaks in the `tabulator-utils.js`, I was able to relax that restriction. Filtering by a `targetClass` is still possible, but if omitted, the collection instead uses all top-level subjects.
- Display of all records in a collection currently relies on that relaxation. On the upside, this means that users are free to mix bibliographical and biographical records in a single collection. On the downside, this means that all field types are available from the column selection menu even if the collection contains only one type of record. Jeroen stated that he only expected to add bibliographical records to collections. I *could* restrict the set of available columns with just a single line of code; however, in that case I would also have to add code that prevents users from adding biographical records, because otherwise users can easily break the interface by adding biographical records anyway.


frontend/vre/main.js
```diff
@@ -27,6 +27,7 @@ import { WelcomeView } from './utils/welcome.view.js';
 GlobalVariables.myCollections = new VRECollections();
 
 // Regular global variables, only visible in this module.
+var unsalientCollections = new VRECollections(); // not currently selected
```
tijmenbaarda, line 30: Not really clear to me what this means; is this all collections except the one that is currently shown (if any)?

tijmenbaarda: I understand now; perhaps change comment to "all collections except currently selected one"?

tijmenbaarda (approved):
Very cool!

The very last commit breaks the whole page, though, or at least in my installation. I get the following error:
afbeelding

When I check out the commit just before that, everything works perfectly.

backend/collect/rdf_models.py
```diff
@@ -1,55 +1,167 @@
...
+}}}}
+insert {{{{
+  {gc_adjusted_count}
+}}}}
+where {{{{
+  {graphs_values}
+  {collection_member}
+  {decrementing_count}
+}}}}
+'''.format
```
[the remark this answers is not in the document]
jgonggrijp: For now, I have addressed this with comments. Still open to other suggestions.

jgonggrijp:
Thanks @tijmenbaarda! I have addressed your comments (I think), including the bug.

PR #261, opened 2025-04-24
Remove records from a collection

### LINKED ISSUE #17: Delete items from collections
After a collection has been created, it should be possible to add items to it later.

jgonggrijp:
I basically implemented record removal by peeking at the logic for record addition. Removal is a bit simpler because you always remove from only one collection at a time. Along the way, there were several minor refactors in order to connect everything.

Removing can be done from a list of records as well as from the data view. In both cases, it looks like a red button that is positioned directly after the control for adding records to collections.

Besides looking at the code, please try out the feature locally. I'm not confident that it looks the way it should look. The remove button is also rather eager; removal cannot be undone (except by looking up the record elsewhere and adding it again), but the user is not given any warning. I'm open to all possible ways to improve the UX.

Closes #17.


frontend/vre/collection/add-to-collection.view.js
```diff
@@ -2,16 +2,16 @@ import _ from 'lodash';
...
+import { AdditionsToCollections } from '../collection/additions-to-collections';
 import { vreChannel } from '../radio.js';
-import collectionTemplate from './collection.view.mustache';
+import additionTemplate from './add-to-collection.view.mustache';
 
 /**
  * View to add a record to a specific collection.
  */
-export var VRECollectionView = View.extend({
+export var AddToCollectionView = View.extend({
```
tijmenbaarda, line 12: Much better name!

tijmenbaarda (approved):
Great! I have no comments on the code.

UX is not ideal indeed, but I would propose to merge this now and look at UX together with the "add to collection" UX, which might also be improved. An undo option would be good to have.

What could maybe still be fixed is that the remove button is not disabled if no selection has been made; this now leads to an exception. (But leaving this for a later PR is also fine by me)

Probably unrelated to this PR: I noticed that when navigating to a certain collection for the second time, the contents are not loaded. Refreshing the SPA fixes the problem. I will create an issue for this.

PR #283, opened 2025-06-16
Add IIIF viewer to the VRE

### LINKED ISSUE #279: Show previews of digitizations and links to the full ones
In the case of IIIF, use the Mirador viewer.

tijmenbaarda:
This adds a `DigitizationView` to show each occurrance of a `DigitizationField` in the right part of the detail modal, and if present, downloads the IIIF manifest and shows a preview image as well as a link to open the manifest in a separate Mirador window.

This partially closes #279 (previews of non-IIIF digitizations still have to be implemented).


tijmenbaarda:
Thanks a lot, this was very helpful and I really appreciate the time you took for this review!

I chose for a slightly different approach, because instead of a IIIF manifest a digitisation field can also simply contain a URL to an external site or an image. I had not yet implemented this, but now I did to make clearer how it works (this requires an updated EDPOP Explorer to see it in action - I will make a separate PR for that). But moving the IIIF manifest fetching code to a separate Backbone model worked very well and the good thing is that the manifest is now cached so the thumbnail shows up more quickly when visiting the record for a second time.

PR #284, opened 2025-06-25
Add originating catalog info to detail page

tijmenbaarda:
Show the name of the original catalog in the detail modal. For this I added an additional block on the right, which saves some vertical space.

I also made the fields table more compact to increase its chance of fitting on the screen...


jgonggrijp (commented):
Code looks good to me, just some minor nitpicks. I also tried out the changes and I like the way it looks.

PR #287, opened 2025-08-06
Feature/fix slow collections

### LINKED ISSUE #254: Collection retrieval API is slow
The retrieval API for collection information, which is used to fill the collections menu, gets slow as more collections are added. After creating five collections it takes over half a second to complete in my installation.

jgonggrijp:
Long-awaited fix #254 and fulfilling a prerequisite for #226. Regarding the former, the result is quite satisfying; the request now finishes in 50-60 ms for a list of 14 collections, down from more than a second.

This is (I think) the first time we convert a writable plain JSON endpoint to JSON-LD, which came with some challenges. Existing tests and existing frontend code had to be adapted, with consequences bleeding out to other modules. A summary of all the changes:

- Added the Django debug toolbar so I could measure and profile the performance of the endpoint.
- Set the DRF default permissions to authenticated or read-only.
- Refactored the project permission logic so parts of it could be reused.
- Rewrote the `/api/collections/` endpoint using `RDFView` and `RDFResourceView` (which requires more code than the old approach based on `RDFModel` and `ModelViewSet`). I used a relatively "rich" context in order to make the final JSON-LD as similar as possible to the original plain JSON.
- Adapted the backend tests to work with the new JSON-LD with context situation.
- Spread over multiple commits for different reasons, I ended up tying up most of the frontend model+collection classes into a single hierarchy by adding more inheritance relationships. `JsonLdModel` now inherits from `APIModel`, `VRECollection` inherits from `JsonLdModel`, and `VRECollections`, `JsonLdNestedCollection` and `JsonLdWithOCCollection` now all inherit from `JsonLdCollection`. I was able to remove some code because of this and to avoid repeating myself, but it mostly served to fix problems.
- Added a variant of `Backbone.sync` that correctly applies the `application/ld+json` mime type when submitting JSON-LD to an endpoint.
- Made `JsonLdModel` and `JsonLdCollection` more capable at parsing different shapes of JSON-LD.
- Adapted `JsonLdModel` and `JsonLdCollection` so they recognize and store the `@context` in JSON-LD and also send it back to the server.
- Added tests for the new logic in the frontend model layer.
- Adapted a backend serializer and frontend model code so that the URI of a project can be used as its identifier. The bare `name` is also still included in the representation, but can no longer be used for lookup via `collection#get` (you can still use `collection#find({name})`).
- Fixed some of the styling around the collection summary editing view that was left broken as a side effect of #253.
- Removed obsolete code: the old `ModelViewSet` endpoint as well as some smaller bits in other modules.

Listing, fetching, creating, updating and deleting collections should all work smoothly, though there is still no frontend UI for deletion (cf #160, #230). That being said, as part of your review, **please do check that everything seems to work**. I rebased the branch on `develop` before submitting this PR for review and may have broken something without realizing it.

Known (possible) open ends:

- While I expect logic, factoring and tests to be mostly in good shape (please do review), I spent little attention on documentation and on what would be the "best" module for a function to live in. It is probably not much effort to add documentation where needed, so please do let me know when you feel this is due.
- There may be other places where something is left broken after #253; I have not investigated this.
- The collection list no longer has a fixed order; it looks very random. I'm open to suggestions on how to improve on this.
- I have not taken any measures to disable the Django debug toolbar in production. Maybe it is always disabled when `DEBUG = False`, but I'm rusty on the matter because it is months ago that I enabled it. I will create a separate issue to document this.
- I have not checked very thoroughly that I removed all obsolete code, so there might still be some dead code lying around.
- Currently most JSON-LD endpoints have their own hardcoded, ad-hoc context. The frontend respects these contexts and includes them when sending back data to the endpoint, but this approach does not feel clean. I think we should have a standalone context document (or maybe a dynamic endpoint) that the endpoints can just refer to with a URL. I will create a separate issue for this as well.

### DIFF
--- a/frontend/vre/utils/jsonld.model.js
+++ b/frontend/vre/utils/jsonld.model.js
@@ -151,8 +151,16 @@ export var JsonLdCollection = APICollection.extend({
         return _.isArray(result) ? result : [result];
     },
     toJSON: function(options) {
+        // By default, individual JsonLdModel instances without an internal
+        // `@context` will look for a context on their containing collection and
+        // copy it into their serialization. When we are serializing the entire
+        // collection, this is redundant. By marking the collection as currently
+        // being serialized, we signal to the invidual models that they need not
+        // copy the collection context.
         this.serializing = true;
         var json = parent(JsonLdCollection.prototype).toJSON.call(this, options);
+        // At this point all calls to `JsonLdModel#toJSON` have finished, so we
+        // can remove the mark again.
         delete this.serializing;
         if (this.context) return {
             '@context': this.context,


tijmenbaarda (approved):
Great, and adding annotations works again, as well. Ready to merge I would say!

PR #286, opened 2025-08-06
Feature/comments

tijmenbaarda:
Add functionality to create, edit and remove annotations to records as a whole (record comments), which closes #19.

Until now, the views for fields and annotations were combined, but this PR separates them, as we discussed before. Fields and annotations are now completely different things. This separation takes place in [c09b766](https://github.com/CentreForDigitalHumanities/EDPOP/pull/286/commits/c09b7664469105c64879a9a2e3a47190f86e0f2f), where some files are splitted and renamed.

Functionality to annotate specific fields (besides commenting also suggesting addition, removal and alteration) will be added later. I expect that large parts of both the API and the web-ui can be reused here.

For now, the API only supports fetching annotations that belong to one single record. In the future, we will need to show the annotations in the table as well (#273), but fetching annotations for single records will remain necessary if we implement #277.

PR #292, opened 2025-08-11
Field annotations

### LINKED ISSUE #269: Adding annotations (RDF version)

### LINKED ISSUE #270: Removing annotations (RDF version)

### LINKED ISSUE #271: Editing annotations (RDF version)

tijmenbaarda:
This pull request closes #269, closes #270 and closes #271.

A field annotation can be made on a specific field occurance - in this case the "originalText" is used as an additional selector. A field annotation may also be made for the field type (such as Title or Contributor); this is especially useful if the field to be annotated does not occur on the original record.

Some restrictions:
- Currently it is only possible to create comments. We will also have to implement suggestions for editing, deletion and additions of field occurances, but we might first want to discuss if this is really necessary.
- It is now only possible to make an annotation on the field itself if it is absent from the original record. We may ask if this is enough for now.
- Field occurances are selected through the original text of the field from the Explorer. If this changes, for example because of a record update, the annotation will be orphaned and will not be visible anymore in the VRE.


jgonggrijp (commented):
This PR appeared very soon after #286 and #288. I am impressed.

The code review took me longer than usual, because the diff included complex old code that moved up (which I realized too late) and because I noticed an issue with the way in which the annotation data are structured. More on that in the comments. Note to self: read the entire diff before starting to write comments. :sweat_smile:

I also tried out the feature and liked the user experience of it. It seems reliable.

In response to the PR description:

> A field annotation can be made on a specific field occurance - in this case the "originalText" is used as an additional selector. A field annotation may also be made for the field type (such as Title or Contributor); this is especially useful if the field to be annotated does not occur on the original record.

To check that I understood this part correctly: the selector of the annotation can either specify both the property and its value, or only the property?

> 
> Some restrictions:
> 
>     * Currently it is only possible to create comments. We will also have to implement suggestions for editing, deletion and additions of field occurances, but we might first want to discuss if this is really necessary.

Good idea to start with something small and ask for feedback before expanding.

> 
>     * It is now only possible to make an annotation on the field itself if it is absent from the original record. We may ask if this is enough for now.

This raised multiple questions for me:

- Do you actually mean "absent" or rather "present but empty"? In the first case, how do the missing fields enter the representation? Perhaps I should already know this, but I am confused.
- Did you mean that the variant where the selector of the annotation only specifies the property is not available when there is a value for the property?
- If the answer to the previous question is yes: what happens if the property gets a value afterwards, when the record is refreshed from its source? Will the comment annotation still be visible in that case?

> 
>     * Field occurances are selected through the original text of the field from the Explorer. If this changes, for example because of a record update, the annotation will be orphaned and will not be visible anymore in the VRE.

This inherent to the nature of these data, so I think we should just accept it and explain it well to the user. However, this could be a reason to enable field-wide comments for fields that already have values. We will need to explain the subtle interaction between the last two bullets to the user as well.

backend/edpop/settings.py
```diff
@@ -103,7 +103,7 @@
 DATABASES = {
     'default': {
         'ENGINE': 'django.db.backends.postgresql_psycopg2',
-        'NAME': 'edpop',
+        'NAME': 'edpop_reserve',
```
jgonggrijp, line 106: This sneaked in again!

tijmenbaarda: Oh no :/

PR #296, opened 2025-08-14
Record reloading and links to other records

### LINKED ISSUE #277: Routing for individual records
Implement URL routing for individual record (by record URI), to make it possible to navigate directly to specific records.

### LINKED ISSUE #278: Add links to related records from detail view (VRE side)
This feature also has to be implemented from the Explorer, see there for the rationale: https://github.com/CentreForDigitalHumanities/edpop-explorer/issues/78

tijmenbaarda:
This PR implements the following related things:

- Add an API view to load single records from the backend (from the triplestore if available, and otherwise through the Explorer)
- Add a `parse` method to `JsonLdModel` so that it can parse incoming data for single subjects
- Allow reloading of records through the detail view
- Add simple routing for showing individual records (close #277)
- Show links to related records from the detail view (now made available through the Explorer - after https://github.com/CentreForDigitalHumanities/edpop-explorer/pull/83 - only for STCN) (close #278)

Feedback I would be particularly interested in:
- I chose to make the RDF of records available through the URIs that the Explorer generates, such as https://edpop.hum.uu.nl/readers/stcn/822911183. That will only match after deployment to the production server, of course, so they are rewritten to `<domain>/readers/stcn/822911183` to make it work wherever. This approach may fail though if readers with other URI roots are added in the future, although that is very hypothetical for now. What do you think?
- I extended the routing in a very basic way: it allows opening of single records in a new window/tab, but it does not add routing to records that are opened in the normal way. It seems to work reliably, but perhaps there are things I did not think about.

### DIFF
--- a/backend/catalogs/api.py
+++ b/backend/catalogs/api.py
@@ -1,4 +1,6 @@
+from django.conf import settings
 from edpop_explorer import ReaderError
+from edpop_explorer.readers.utils import get_record_by_uri
 from typing import Optional
 
 from rdf.renderers import TurtleRenderer, JsonLdRenderer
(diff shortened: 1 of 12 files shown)


tijmenbaarda:
Thanks! I created an issue in the Explorer repository about being able to change the root IRIs.

Issue #124, opened 2022-09-22
Convert the entire application to RDF


jgonggrijp:
We're going to do more data sources, more querying, more linking and more annotating. We can, and should, reuse infrastructure built for READ-IT. Depends on outfactoring the relevant components from that project.

Enables (a possible flavor of) #94, #78, #67, #34, #19. Supersedes #79, #47.

Issue #294, opened 2025-08-13
RDF structure of annotations is incorrect


tijmenbaarda:
> I didn't notice this when reading `backend/annotations/api.py`, but this structuring of the annotation seems a bit off.
> 
> Firstly, `selectField` and `selectOriginalText` don't seem to appear anywhere in the EDPOP collection ontology. Perhaps you meant `edpopcol:property` and `edpopcol:object`, respectively.
> 
> Secondly, you seem to have conflated three resources into one. Normally, an `oa:Annotation` has a body, a target and a motivation. The target then has a source and a selector. The selector, in turn, has properties that identify a part of the source. In this snippet of code, you seem to have used `oa:hasTarget` to identify what is actually the `oa:hasSource` of the target, while the same object also contains properties of the selector.
> 
> On the one hand, I realize it is convenient to avoid nested objects in model attributes. On the other hand, part of the selling point of RDF is portability, so we must ensure that our data are well-formed.
> 
> There are a few ways in which we could have both:
> 
> - The easiest would be a hypothetical notation in a JSON-LD context that just lets us fold nested properties. I was not able to determine whether this is possible, so I asked in https://github.com/json-ld/json-ld.org/issues/877.
> - The next easiest would be to define `parse` and `toJSON` methods for the `Annotation` class that convert between true JSON-LD and a folded representation on the fly. This is a bit trickier because the `parse` method will need to take multiple possible notations into account.
> - A variant of the previous option is to make a separation between a model type that contains true JSON-LD and a model type that contains a folded representation of the former.
> - In READ-IT, we took the previous option, with the simplifying circumstance that we involved a third-party multi-format RDF parser that normalized all notations to flattened and expanded JSON-LD. This would be our most sustainable solution long-term, but unfortunately I still haven't outfactored that functionality.
> 
> Otherwise:
> 
> - Leave as is, for the time being. I'm fine with merging this to `develop` for demonstration purposes, but I would not want to deploy this in production, because in that case we would burden ourselves with a data migration in the future.
> - Turn `Annotation` into a model with nested attributes. Probably painful, but maybe workable.

_Originally posted by @jgonggrijp in https://github.com/CentreForDigitalHumanities/EDPOP/pull/292#discussion_r2269765607_

PR #311, opened 2025-09-30
Add script to export and import legacy data

tijmenbaarda:
Add a script to export legacy data to JSON and to import it to the triplestore. The scripts are not particularly beautiful and not all functionality that is shared with API views has been outfactored, but the scripts are supposed to be used only for the existing data. They may go away after this is done.

Usage: `python manage.py export_legacy_data` and `python manage.py import_legacy_data <file> <project-IRI>`.

By using JSON as an intermediate format, there is no need to copy database contents around servers.

In the current state, the import script is only meant to be run on the acceptation server. After fixing #294, the script will need to be slightly adapted and the script can be run on the production server.

This script does not copy user data, because we do not want that on the acceptation server. For copying usernames and password hashes, I suggest using a table dump and restore directly on the server.

### DIFF
--- a/backend/triplestore/management/commands/import_legacy_data.py
+++ b/backend/triplestore/management/commands/import_legacy_data.py
@@ -1,0 +1,168 @@
+import datetime
+import json
+import time
+
+from django.conf import settings
+from django.core.management import BaseCommand
+from rdflib import Graph, URIRef, Literal, RDF, DCTERMS
+
+from edpop_explorer.readers import HPBReader
+from edpop_explorer import NotFoundError
+
+from accounts.utils import IMPORT_USER_URIREF
+from annotations.api import create_annotation_subject_node, ANNOTATION_GRAPH_IDENTIFIER
+from catalogs.triplestore import save_to_triplestore, remove_from_triplestore
+from catalogs.utils import record_exists
+from collect.blank_record import BlankRecordReader, create_blank_record
+from collect.rdf_models import EDPOPCollection
+from collect.utils import collection_uri, collection_graph, collection_exists
+from triplestore.constants import AS, EDPOPCOL, OA
+from triplestore.utils import triples_to_quads
+
+RecordMapping = dict[str, str]
+
+glossary_mapping = {  # There are more entries, but these are the only ones that are in use
+    "Chapbook": "https://popular-print-glossary.sites.uu.nl/glossary/chapbook/",
+    "Almanac": "https://popular-print-glossary.sites.uu.nl/glossary/almanac/",
+    "Abecedarium": "https://popular-print-glossary.sites.uu.nl/glossary/abecedarium/",
+    "Catechism primer": "https://popular-print-glossary.sites.uu.nl/glossary/catechism-primer/",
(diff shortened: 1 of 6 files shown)


backend/triplestore/management/commands/import_legacy_data.py
```diff
@@ -0,0 +1,168 @@
...
+        # Add as record comment - may be changed to field comment manually
+        body = Literal(f"{annotation_key}: {annotation_value}")
+        motivation = OA.commenting
+    subject_node = create_annotation_subject_node()
+    as_published = Literal(datetime.datetime.now())
+    dcterms_creator = IMPORT_USER_URIREF
+    graph = Graph(identifier=ANNOTATION_GRAPH_IDENTIFIER)
+    triples = [
+        (subject_node, RDF.type, EDPOPCOL.Annotation),
+        (subject_node, OA.hasTarget, URIRef(record_iri)),
```
jgonggrijp, line 138:
```py
from rdflib import BNode
from triplestore.utils import replace_blank_node
```

```suggestion
    subject_node = create_annotation_subject_node()
    target_node = replace_blank_node(BNode())
    as_published = Literal(datetime.datetime.now())
    dcterms_creator = IMPORT_USER_URIREF
    graph = Graph(identifier=ANNOTATION_GRAPH_IDENTIFIER)
    triples = [
        (subject_node, RDF.type, EDPOPCOL.Annotation),
        (subject_node, OA.hasTarget, target_node),
        (target_node, OA.hasSource, URIRef(record_iri)),
```

frontend/vre/annotation/annotation.edit.view.js
```diff
@@ -43,7 +43,7 @@ export var AnnotationEditView = View.extend({
         if (this.model.getAnnotationType() === 'tag') {
             this.$('select').select2('destroy');
             this.$el.html(this.glossaryTemplate({
-                choices: glossary.toJSON(),
+                choices: glossary.models.map(function(model) {return model.toJSON();}),
```
jgonggrijp, line 46: Ah, you found that bug, too. I fixed it in a different way in #313.

jgonggrijp (changes requested):
Looks good to me, just a minor change in order to ensure that the migrated annotations will immediately work after #313.

> Add a script to export legacy data to JSON and to import it to the triplestore. The scripts are not particularly beautiful and not all functionality that is shared with API views has been outfactored, but the scripts are supposed to be used only for the existing data.

I can live with that!

> They may go away after this is done.

Better retain the ugly scripts! Just in case someone in the future wants to run a prior version of the VRE and then migrate data again.

> 
> Usage: `python manage.py export_legacy_data` and `python manage.py import_legacy_data <file> <project-IRI>`.
> 
> By using JSON as an intermediate format, there is no need to copy database contents around servers.

Smart. :+1: 

> 
> In the current state, the import script is only meant to be run on the acceptation server. After fixing #294, the script will need to be slightly adapted and the script can be run on the production server.

I already suggested that adaptation below, because my fix for #294 is ready (#313).

> 
> This script does not copy user data, because we do not want that on the acceptation server. For copying usernames and password hashes, I suggest using a table dump and restore directly on the server.

Finy by me.

> 
> I believe that this will finally close #124.

Agreed!

PR #313, opened 2025-09-30
Restructure annotations

### LINKED ISSUE #294: RDF structure of annotations is incorrect

jgonggrijp:
This closes #294, which I was able to do faster than I expected.

I previously described an approach for the frontend model based on `get`/`set` accessors. I changed my mind about that, because that approach would not trigger the conventional `change` events unless I would trigger them manually from the accessors. It also meant that idiomatic `annotation.get()` and `annotation.set()` would still be cumbersome. Instead, I went with the `parse`/`toJSON` approach that I already described in #294, so that everything could stay conventional and idiomatic.

The approach I took has a different disadvantage, which I consider the lesser evil. The model stores both the faithful, nested JSON-LD representation of the annotation, in order to prevent information loss, as well as a partial flattened representation for convenience. Where information is duplicated, users of the model class are expected to always access and edit the flattened copy of the property rather than the nested copy. This *could* be a footgun, if a user is knowledgeable about the web annotation vocabulary and JSON-LD but not about the `Annotation` model class. It also introduces some inconsistency in the naming of the attributes,  as you will see when you read the JsDoc comments in the `frontend/vre/annotation/annotation.model.js` module. I'm open to suggestions for modifications on this front.

Other changes:

- rewrote of the backend API, not only to reflect the structural changes but also to ingest proper JSON-LD instead of plain JSON
- avoiding the word "target" wherever `oa:hasSource` is intended, including in one of the backend API routes (with one exception for `Annotations#target`; this could perhaps be renamed to `#record`)
- removed or simplified a lot of code in the frontend
- renamed the selector fields to reflect the proposed changes in CentreForDigitalHumanities/edpop-collection-ontology#7
- fixed a previously unnoticed bug that I caused in #287, where the tag editor could no longer list the genres from the glossary

### DIFF
--- a/backend/annotations/api.py
+++ b/backend/annotations/api.py
@@ -1,138 +1,171 @@
+import datetime
 import uuid
 from django.conf import settings
-from django.http.response import HttpResponse
 from rest_framework.views import Request
-from rest_framework.parsers import JSONParser
-from rest_framework.decorators import api_view, renderer_classes
 from rest_framework.response import Response
+from rest_framework.exceptions import ValidationError
 from rdflib import URIRef, Graph, Literal, DCTERMS, RDFS, RDF
-import datetime
-
-from triplestore.constants import EDPOPREC, AS, EDPOPCOL
-from rdf.views import RDFView
+from rdf.views import RDFView, graph_from_request
 from rdf.utils import graph_from_triples
 from rdf.renderers import TurtleRenderer, JsonLdRenderer
+
 from accounts.utils import user_to_uriref
-from triplestore.constants import OA, AS
-from triplestore.utils import triples_to_quads, sparql_multivalues
+from triplestore.constants import EDPOPREC, OA, AS, EDPOPCOL
+from triplestore.utils import (
+    replace_blank_nodes_in_triples,
+    replace_node,
+    sparql_multivalues,
+    triples_to_quads,
+)
 
 ANNOTATION_GRAPH_URI = settings.RDF_NAMESPACE_ROOT + "annotations/"
 ANNOTATION_GRAPH_IDENTIFIER = URIRef(ANNOTATION_GRAPH_URI)
 
 RDF_ANNOTATION_ROOT = settings.RDF_NAMESPACE_ROOT + "annotations/"
 
+NS = {
+    'rdfs': RDFS,
+    'edpoprec': EDPOPREC,
+    'edpopcol': EDPOPCOL,
+    'oa': OA,
+    'as': AS,
+    'dcterms': DCTERMS,
+}
+
 JSON_LD_CONTEXT = {
-    'rdfs': str(RDFS),
-    'edpoprec': str(EDPOPREC),
-    'edpopcol': str(EDPOPCOL),
-    'oa': str(OA),
-    'as': str(AS),
-    'dcterms': str(DCTERMS),
+    prefix: str(namespace)
+    for prefix, namespace in NS.items()
 }
 
 
-# Argument: target_uris
-collection_records_query = '''
-construct {{
-  ?s ?p ?o .
-}}
-where {{
-  values ?annotation {{ {target_uris} }}
-  graph ?annotations {{
-    ?s ?p ?o.
-    ?s <http://www.w3.org/ns/oa#hasTarget> ?annotation.
-  }}
-}}
-'''.format
+record_annotations_query = '''
+construct {
+  ?a ?pa ?oa .
+  ?t ?pt ?ot .
+  ?s ?ps ?os .
+}
+where {
+  graph ?annotations {
+    ?a ?pa ?oa ;
+       oa:hasTarget ?t .
+    ?t ?pt ?ot ;
+       oa:hasSource ?record .
+    optional {
+      ?t oa:hasSelector ?s .
+      ?s ?ps ?os .
+    }
+  }
+}
+'''
 
 delete_annotation_update = '''
 delete {
   graph ?annotations {
-    ?annotation ?p ?o.
+    ?annotation ?pa ?oa .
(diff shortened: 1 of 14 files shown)


backend/annotations/api.py
```diff
@@ -1,138 +1,171 @@
...
-        quads = list(triples_to_quads(triples, graph))
+        quads = list(triples_to_quads(clean_triples, graph))
         # Get the existing graph from Blazegraph
         store = settings.RDFLIB_STORE
         store.addN(quads)
         store.commit()
-        graph = graph_from_triples(triples)
-        return Response(graph)
+        response_graph = graph_from_triples(clean_triples)
+        return Response(response_graph)
```
tijmenbaarda, line 165: Beautiful! Much cleaner than before too.

backend/annotations/api.py
```diff
@@ -1,138 +1,171 @@
...
+        ?target oa:hasSelector ?selector .
+        ?selector ?ps ?os .
+      }
+    }
   }
 }
 '''
 
-delete_annotation_body_update = '''
+update_annotation_body = '''
```
tijmenbaarda, line 84: Ah, I didn't know that delete and insert can be done in one query :)

frontend/vre/annotation/annotation.model.js
```diff
@@ -1,71 +1,134 @@
...
+        if (flatField) selector['edpopcol:field'] = {'@id': flatField};
+        if (flatText) selector['edpopcol:originalText'] = flatText;
+        if (flatField || flatText) var flatSelector = true;
+        if (flatSource) target['oa:hasSource'] = {'@id': flatSource};
+        if (flatSelector) target['oa:hasSelector'] = selector;
+        if (flatSource || flatSelector) jsonld['oa:hasTarget'] = target;
+        if (flatTagURL) jsonld['oa:hasBody'] = {'@id': flatTagURL};
+        if (flatMotivation) jsonld['oa:motivatedBy'] = {'@id': flatMotivation};
+        _.each(this.flatFields, stripAttribute(jsonld));
+        return jsonld;
```
tijmenbaarda, line 104: I like this solution a lot!

frontend/vre/annotation/annotation.model.js
```diff
@@ -1,71 +1,134 @@
...
+ * Backbone.Model representation of a Web Annotation to an EDPOP record.
+ *
+ * This model stores and synchronizes an annotation in compacted JSON-LD format.
+ * The Web Annotation data model is highly nested, which is inconvenient in a
+ * presentation and editing context. To remedy this, nested properties of
+ * interest are copied to the root `attributes` object during parsing. They are
+ * re-nested before being sent to the backend. Client code should read and edit
+ * these root-level attributes, rather than their standard nested paths (in parentheses):
+ *
+ * motivation            ([oa:motivatedBy][@id])
```
tijmenbaarda, line 25: Is there a reason why you didn't keep `oa:motivatedBy` here?

tijmenbaarda: By the way, very nice to have this documentation comment.

tijmenbaarda (approved):
Wonderful! I have one minor comment but no request for changes.

frontend/vre/annotation/annotation.model.js
jgonggrijp: Glad you like the documentation. Yes, the reason I don't reuse `oa:motivatedBy` is because that would be ambiguous (it is always present on the annotation itself, so if it is a string, is that because we flattened it or because the type changed from IRI to literal?). The flattened attributes in which I preserved the `oa:` prefixed name all come from deeper in the structure. I used `tagURL` instead of reusing `oa:hasBody` for the same reason.

PR #324, opened 2025-11-27
Project-bound annotations

### LINKED ISSUE #315: Associate each annotation with a project
In a distant past, before the adoption of RDF, the project selection menu was called an "annotation context" menu. Every annotation was associated with the project that was selected in that menu.

This association should return, i.e., when an annotation is created, it should store the project to which it applies as its target's `oa:hasScope` (or `as:context`). This will require a small sophistication to the `flatFields`, `parse` and `toJSON` properties of `frontend/vre/annotation/annotation.model.js:Annotation.prototype`. In addition, when the frontend creates an annotation, it will have to `set()` the currently selected project on the new instance, similar to how collections are already associated with the current project. Furthermore, the migration script should add this information to migrated annotations.

To summarize the above:

- [x] Add project context to the anticipated fields in the backend `POST` endpoint for creating new annotations.
- [x] Adapt `flatFields`, `parse` and `toJSON` in the frontend `Annotation` model to incorporate `oa:hasTarget/oa:hasScope`.
- [x] In the frontend code that creates new annotations, request and store the currently selected project in the annotation, similarly to how this is done for collections.
- [x] In the data migration script (#226), add the triples that associate each annotation with its project.

jgonggrijp:
This closes #315. It was easier than I expected.

Frontend and backend cooperate to ensure that each annotation is associated with a project. The backend also enforces that the user must have write access to the project. The `import_legacy_data` script adds a triple to backport the association.

As part of your review, please try re-running the legacy import script on the acceptance server and check that the annotations come out bound to the project. The easiest way to verify this is probably to visit the frontend and alt-click on one of the annotations. That will log three or four views to the console, from innermost to outermost. The innermost model's `.model.attributes` should contain both an `as:context` and a `context` property with the full IRI of the project.

### DIFF
--- a/backend/annotations/conftest.py
+++ b/backend/annotations/conftest.py
@@ -1,0 +1,6 @@
+import pytest
+
+
+@pytest.fixture
+def django_test_user(django_user_model):
+    return django_user_model.objects.create_user(username='tester', password='secret')
--- a/backend/annotations/conftest.py
+++ b/backend/annotations/conftest.py
@@ -1,6 +1,19 @@
 import pytest
 
+from projects.models import Project
+
 
 @pytest.fixture
 def django_test_user(django_user_model):
     return django_user_model.objects.create_user(username='tester', password='secret')
+
+
+@pytest.fixture()
+def project(db, django_test_user):
+    project = Project.objects.create(
+        name='project',
+        display_name='Project',
+    )
+    project.users.add(django_test_user)
+    project.save()
+    return project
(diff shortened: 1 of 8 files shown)


tijmenbaarda:
Fantastic news and wonderful to see that it involves such a limited number of changes. I hope to get to it today; otherwise on Monday.

backend/annotations/conftest.py
```diff
@@ -0,0 +1,19 @@
+import pytest
+
+from projects.models import Project
+
+
+@pytest.fixture
+def django_test_user(django_user_model):
+    return django_user_model.objects.create_user(username='tester', password='secret')
```
tijmenbaarda, line 8: Nice to have these here

tijmenbaarda (approved):
Great, this works all as expected. As requested, I have tested the import script on the acceptation server as well, which results in correct annotations.

PR #326, opened 2025-11-28
Removed deprecated convert_legacy_data management command

Neafire:
i removed the redundent code


jgonggrijp (changes requested):
@Neafire Thank you for trying to contribute to an open source project. You are on the right track, but the file you deleted is not the only one that is obsolete. Please trace down the other code that is also obsolete and remove that as well.

Neafire:
i removed the convert_legacy_data folder containing the tests for the legacy code now.

I am new to this so sorry i missed it.

jgonggrijp (commented):
No need to apologize, everyone starts out new.

You have now removed most of the obsolete code, but there is still a little bit more. You can find individual functions from other modules, that are imported in the deleted modules but not anywhere else.

Ideally, you run the tests before and after deleting the remaining code. The easiest way to do this is using Docker (or Podman):

```sh
# get everything started and running (will take long the first time)
docker compose up backend
# run tests the first time
docker compose exec backend pytest
# remove the remaining obsolete code
# run tests again
docker compose exec backend pytest
# also sanity-check the application
docker compose exec backend python manage.py check
# shut down the application
docker compose down
```

The tests might output warnings but should still pass without errors.

Please try it out and see how far you get.

Neafire:
Removed `find_subject_by_class` and `objects_to_graph`, which were only referenced
by the deprecated convert_legacy_data code path.

I ran the full test suite before and after the removal, all 75 tests passed
(with the same 12 expected warnings). `manage.py check` also reported no issues.

As a small bonus, removing the unused code shaved about 1 second off the test runtime.

Thanks for the opportunity to learn from this

jgonggrijp (changes requested):
Well done. I think I was able to find more functions that were exclusive to the convert command yesterday. Could you dig a bit deeper still? Some of those functions might have tests that should be removed as well.

Neafire:
@jgonggrijp I did a deeper pass and searched the entire backend for any remaining
functions or tests tied to the old convert_legacy_data code path.

I ran:

 grep -R "convert" -n backend/
 grep -R "legacy" -n backend/
searched for all private helper names (_record_to_graph, _field_to_graph, etc.)
verified no leftover test modules referenced the removed functionality

All occurrences I found were either:
- unrelated comments (e.g. “convert it” in sru_query.py), or
- part of active utilities such as replace_blank_node, or
- the still-supported import/export legacy commands.

No remaining code or tests appear to be linked to the old convert_legacy_data
implementation.

If you know specific names or files you suspect are still present, I’m happy
to dig further, but based on the full-text search, the convert command’s
helpers and tests are now fully removed.

jgonggrijp (changes requested):
The way to do this, is to trace the imports in each removed module and then check that the function isn't used elsewhere. You missed these:

- [ ] `ObjectURIs`
- [ ] `union_graphs`
- [ ] `import_ontology` and `download_ontology` (the latter is only used in the former)

Neafire:
I’ve gone through the code again and removed the last bits that were only used
by the old convert_legacy_data flow:

- ObjectURIs
- union_graphs (and its tests)
- import_ontology / download_ontology (and their test)

I checked each of them to make sure nothing else was using them.

Ran the full test suite before and after (3 tests gone now) and everything still
passes. Also ran `manage.py check` and that came back clean.

Thanks for pointing me in the right direction I learned learned a lot doing this.

jgonggrijp:
Congratulations with completing this until the end and thank you for contributing in the open source spirit.

Issue #309, opened 2025-09-23
After adding a record to a collection, it does not become visible in target collection if already loaded


tijmenbaarda:
To make the record visible in the target collection, it is necessary to reload the VRE.

PR #335, opened 2025-12-18
Update live

### LINKED ISSUE #123: Use backbone-fractal and backbone.radio where applicable

### LINKED ISSUE #130: Some frontend views are leaky

### LINKED ISSUE #131: Render-placement antipattern

jgonggrijp:
This fixes the long-standing but late-documented issue where added records would not be visible in previously cached collections (#309).

I employed "spooky action at a distance" through the radio channel. Since I was at it, I also eliminated the last usage of `GlobalVariables`.

After merging this, I think we can also close the following issues. I am not 100% sure all instances of those issues have been cleared up, but I am quite sure that they are now so rare that they don't warrant open issues anymore. We can still refer to the tickets in commit messages after closing them.

Close #123, close #130, close #131.

As part of your review, I invite you to try it out locally.

### DIFF
--- a/frontend/vre/collection/add-to-collection.view.js
+++ b/frontend/vre/collection/add-to-collection.view.js
@@ -55,7 +55,7 @@ export var AddToCollectionView = View.extend({
         var selected_collections = this.$('select').val();
         if (!selected_collections.length) return;
         var records_and_collections = new AdditionsToCollections({
-            'records': selected_records,
+            'records': _.map(selected_records, 'id'),
             'collections': selected_collections,
         });
         records_and_collections.save().then(
--- a/frontend/vre/collection/add-to-collection.view.js
+++ b/frontend/vre/collection/add-to-collection.view.js
@@ -58,7 +58,7 @@ export var AddToCollectionView = View.extend({
             'records': _.map(selected_records, 'id'),
             'collections': selected_collections,
         });
-        records_and_collections.save().then(
+        records_and_collections.save(null, {records: selected_records}).then(
             this.showSuccess.bind(this),
             this.showError.bind(this),
         );
--- a/frontend/vre/collection/add-to-collection.view.js
+++ b/frontend/vre/collection/add-to-collection.view.js
@@ -77,7 +77,7 @@ export var AddToCollectionView = View.extend({
     },
     showAlert: function(level, message) {
         var alert = new AlertView({level: level, message: message});
-        alert.render().$el.prependTo(this.el);
+        alert.$el.prependTo(this.el);
         alert.animate('remove');
     },
 });
--- a/frontend/vre/collection/additions-to-collections.js
+++ b/frontend/vre/collection/additions-to-collections.js
@@ -1,5 +1,16 @@
 import Backbone from 'backbone';
 
+import { vreChannel } from '../radio.js';
+
 export var AdditionsToCollections = Backbone.Model.extend({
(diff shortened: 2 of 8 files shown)


tijmenbaarda (approved):
Beautiful! I tried it out and it looks like it is working well. Happy that the other issues could be closed as well!

PR #339, opened 2026-01-13
Record ordering fixes

### LINKED ISSUE #312: Moving left and right in modal does not follow table order anymore

### LINKED ISSUE #338: Keep original order of catalogue results
Due to a bug, the catalogue results are ordered by title/name by default. I am working on a solution.

tijmenbaarda:
This PR fixes some problems related to the ordering of records:

* Close #312: follow the record order according to the table when going back and forth in the modal
* Close #338: keep the original order of catalogue records by default (this was already supposed to work, but due to a bug it didn't)
* Add an item to the context menu that lets you restore the order of the table
* As a bonus, this PR adds highlighting of the currently selected record

### DIFF
--- a/frontend/vre/css/style.css
+++ b/frontend/vre/css/style.css
@@ -71,6 +71,15 @@ form.form-inline.inline-editor {
     z-index: 1060;
 }
 
+.highlighted {
+    font-weight: bold;
+}
+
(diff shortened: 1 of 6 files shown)


jgonggrijp (approved):
Love it!

PR #340, opened 2026-01-13
Show loading status and number of records in collections

### LINKED ISSUE #331: Show number of records in collection view

tijmenbaarda:
Close #331.

### DIFF
--- a/frontend/vre/collection/browse-collection.view.js
+++ b/frontend/vre/collection/browse-collection.view.js
@@ -32,6 +32,8 @@ export var BrowseCollectionView = CompositeView.extend({
         });
         overlay.listenTo(editor, 'submit reset', overlay.uncover);
 
+        this.listenTo(this.model.records, 'complete', this.render);
+
         this.render().listenTo(this.model, 'change', this.render);
     },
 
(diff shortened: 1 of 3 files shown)


frontend/vre/collection/browse-collection.view.js
```diff
@@ -45,7 +47,10 @@ export var BrowseCollectionView = CompositeView.extend({
     },
 
     renderContainer: function() {
-        this.$el.html(this.template(this.model.attributes));
+        this.$el.html(this.template(Object.assign(this.model.attributes, {
+            numberOfRecords: this.collection.length,
+            loaded: this.model.loaded,
+        })));
```
jgonggrijp: This will actually add `numberOfRecords` and `loaded` properties to `this.model.attributes`, which means they will also be transmitted to the server when `this.model` is saved. Is this intentional? Otherwise, swap the order of the arguments to `Object.assign`.

Side note: since we're using Underscore/Lodash anyway, you could use `_.assign` or `_.extend` instead of `Object.assign`. Saves a few bytes and is more backwards compatible.

tijmenbaarda: Oh no, I thought `Object.assign` was making a copy... thinking about that that does not make sense. Will change and will use `_.assign`.

Issue #319, opened 2025-11-14
Restrict collection selection to collections that belong to the currently selected project


jgonggrijp:
Right now, users always see all collections they have access to. To isolate projects better from each other, the collection menu and the add-to-collection picker should be restricted to those collections that belong to the currently selected project.

Issue #317, opened 2025-11-14
Filter /api/record-annotations/ not only by record but also by project


jgonggrijp:
This route, and its corresponding view class, should be updated so that the returned annotations are not only restricted to a single record, but also to a single project.

Issue #316, opened 2025-11-14
Fetch only annotations belonging to the currently selected project


jgonggrijp:
The frontend model methods `Annotations#initialize` and `Record#getAnnotations` currently fetch all annotations associated with the salient record. This should be further narrowed down to annotations that belong to the currently selected project.

PR #342, opened 2026-01-21
Filter annotations based on currently selected project

jgonggrijp:
I started this branch a month ago, but I got stuck on a failing test right after I started and that was just before the Christmas holidays. Fortunately, a break does wonders to get unstuck.

I modified the `/api/record-annotations/` endpoint to accept and expect a new `?project=<URIEncoded project URI>` query parameter (#317). Initially, I wanted to go with a path component, just like with the record URI that is already passed to this endpoint, but Django's router won't parse this correctly. It seems that Django's middleware *first* unencodes URI components and *then* parses the endpoint, causing even escaped slashes to be processed.

I then modified the `Annotations` collection class in the frontend to use the new query parameter (#316). As I was testing this, I discovered a problem with another frontend test that I fixed in the last commit of this branch.

The tests pass and the application correctly filters annotations based on the currently selected project. Due to #319 not being implemented yet, this can lead to confusing behavior in local development: if you select one project, open a collection, open a record, make an annotation, close the record, select another project and open the same record again, you will still see your annotation. You can make the cached annotation disappear by refreshing the page. If you then switch back to the first project and open the record again, your annotation will not appear, until you refresh the page again. Fortunately, this should not be an issue in production because there is only one project at this time. I expect the bug to disappear when we implement #319.

### DIFF
--- a/backend/annotations/api_test.py
+++ b/backend/annotations/api_test.py
@@ -16,7 +16,7 @@ def create_annotation(client, target, body, django_test_user, project) -> str:
         "@context": JSON_LD_CONTEXT,
         "oa:hasTarget": target,
         "oa:hasBody": body,
-        "as:context": project.uri,
+        "as:context": {'@id': project.uri},
     }
     client.force_login(django_test_user)
     response = client.post('/api/annotation/', data, content_type='application/ld+json')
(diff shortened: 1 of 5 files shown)


backend/annotations/api_test.py
```diff
@@ -16,7 +16,7 @@ def create_annotation(client, target, body, django_test_user, project) -> str:
         "@context": JSON_LD_CONTEXT,
         "oa:hasTarget": target,
         "oa:hasBody": body,
-        "as:context": project.uri,
+        "as:context": {'@id': project.uri},
```
jgonggrijp, line 19: This line is what got me stuck just before the holiday. It was an oversight that I made in #324. It took me a lot of debugging to figure out what I did wrong!

tijmenbaarda (approved):
Great; this works correctly with the caveats you mentioned!

PR #343, opened 2026-01-22
Upgrade selected dependencies

tijmenbaarda:
Update selected backend and frontend dependencies to solve security issues.


backend/requirements.txt
```diff
@@ -1,8 +1,8 @@
 #
-# This file is autogenerated by pip-compile with Python 3.9
+# This file is autogenerated by pip-compile with Python 3.13
```
jgonggrijp: Is this the version we're running in production? In that case I should probably update the docker image.

jgonggrijp (approved):

backend/requirements.txt
tijmenbaarda: Oh! I didn't realize I am working with a different Python version. No, production has 3.9. Hold on, I'll redo this with Python 3.9...

PR #344, opened 2026-01-23
Add tag (glossary item) to table view

### LINKED ISSUE #306: Add column for glossary tags to tabulator view
This would enable users to sort and filter based on glossary categories.

### LINKED ISSUE #341: Add "originating catalogue" column to table view for collections

tijmenbaarda:
Close #306.

In the current implementation, all annotations are loaded one by one, which is inefficient. However, loading all annotations at once requires some design choices that I would like to discuss before starting work on them, and in the meantime I would like to work on some other issues. @jgonggrijp: if you think that we can better fix loading all annotations at once before merging this, I understand.

Also close #341 (add "original catalogue" column).

### DIFF
--- a/frontend/vre/record/record.list.view.js
+++ b/frontend/vre/record/record.list.view.js
@@ -66,7 +66,7 @@ export var RecordListView = Backbone.View.extend({
             height: "calc(100vh - 360px)", // set height to table approximately to what is left of viewport height
             data: initialData,
             autoColumns: true,
-            autoColumnsDefinitions: (autodetected) => {return adjustDefinitions(autodetected, this.recordClass)},
+            autoColumnsDefinitions: (autodetected) => {return adjustDefinitions(autodetected, this.recordClass, this.type)},
             layout: "fitColumns",
             initialSort: getDefaultSort(this.recordClass, this.type),
             resizableColumnFit: true,
--- a/frontend/vre/record/record.list.view.js
+++ b/frontend/vre/record/record.list.view.js
@@ -57,6 +57,7 @@ export var RecordListView = Backbone.View.extend({
         const data = this.collection.toTabularData();
         if (this.table === null) return this.createTable(data);
         this.table.replaceData(data);
+        this.connectToData();
         return this;
     },
 
@@ -107,11 +108,24 @@ export var RecordListView = Backbone.View.extend({
             const model = cell.getRow().getData().model;
             vreChannel.trigger('displayRecord', model);
         });
+        this.table.on("tableBuilt", this.connectToData.bind(this));
         vreChannel.on('highlightRecord', this.highlightRecord.bind(this));
         vreChannel.on('unhighlightRecord', this.unhighlightRecord.bind(this));
         vreChannel.reply('getNextRecord', this.getNextRecord.bind(this));
         vreChannel.reply('getPreviousRecord', this.getPreviousRecord.bind(this));
         return this;
+    },
+
+    connectToData: function() {
+        var rows = this.table.getRows();
+        for (let row of rows) {
+            var record = row.getData().model;
+            record.on("annotations:loaded", function(model) {
+                var tabularData = model.toTabularData();
+                this.table.updateData([tabularData]);
+            }.bind(this));
+            record.getAnnotations();
+        }
     },
 
     removeTable: function() {
--- a/frontend/vre/record/record.list.view.js
+++ b/frontend/vre/record/record.list.view.js
@@ -49,15 +49,17 @@ export var RecordListView = Backbone.View.extend({
 
     initialize: function(options) {
         _.assign(this, _.pick(options, ['recordClass', 'type']));
+        this.listenTo(this.collection, 'annotations:loaded', this.insertRowAnnotations);
         this.render().listenTo(this.collection, 'update', this.render);
     },
 
     render: function() {
         if (this.collection.length === 0) return this.removeTable();
         const data = this.collection.toTabularData();
+        _.invokeMap(this.collection.models, 'getAnnotations');
         if (this.table === null) return this.createTable(data);
         this.table.replaceData(data);
-        this.connectToData();
+        // TODO: replace by _.invoke when switching to Underscore
         return this;
     },
 
@@ -108,7 +110,6 @@ export var RecordListView = Backbone.View.extend({
             const model = cell.getRow().getData().model;
             vreChannel.trigger('displayRecord', model);
         });
-        this.table.on("tableBuilt", this.connectToData.bind(this));
         vreChannel.on('highlightRecord', this.highlightRecord.bind(this));
         vreChannel.on('unhighlightRecord', this.unhighlightRecord.bind(this));
         vreChannel.reply('getNextRecord', this.getNextRecord.bind(this));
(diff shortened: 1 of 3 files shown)


frontend/vre/record/record.list.view.js
```diff
@@ -107,13 +108,26 @@ export var RecordListView = Backbone.View.extend({
...
         return this;
     },
 
+    connectToData: function() {
+        var rows = this.table.getRows();
+        for (let row of rows) {
+            var record = row.getData().model;
+            record.on("annotations:loaded", function(model) {
+                var tabularData = model.toTabularData();
+                this.table.updateData([tabularData]);
```
tijmenbaarda: The `updateData()` method replaces data in the table based on the `id` property.

frontend/vre/utils/tabulator-utils.js
```diff
@@ -44,7 +44,8 @@ export const columnChooseMenu = function(){
         let label = document.createElement("span");
         let title = document.createElement("span");
 
-        title.textContent = " " + (definition.titleInMenu || definition.title);
+        // For textContent, prefer the headerTooltip, which is the text representation in case the field
```
jgonggrijp: Unfinished sentence in this comment?

frontend/vre/record/record.list.view.js
```diff
@@ -107,13 +108,26 @@ export var RecordListView = Backbone.View.extend({
...
+            }.bind(this));
```
jgonggrijp: If you outfactor this anonymous handler `function(model)` to a regular method of the view, say `insertRowAnnotations`, ...

frontend/vre/record/record.list.view.js
```diff
@@ -57,6 +57,7 @@ export var RecordListView = Backbone.View.extend({
         const data = this.collection.toTabularData();
         if (this.table === null) return this.createTable(data);
         this.table.replaceData(data);
+        this.connectToData();
```
jgonggrijp: ... you won't need the `connectToData` method, and you can just write this here:


```suggestion
        this.listenTo(this.collection, 'annotations:loaded', this.insertRowAnnotations);
        // TODO: replace by _.invoke when switching to Underscore
        _.invokeMap(this.collection, 'getAnnotations');
```

because `_.map(this.table.getRows(), 'model')` is an array with the same contents as `this.collection.models` and because `this.listenTo` binds the handler to `this` by default.

frontend/vre/record/record.list.view.js
```diff
@@ -107,13 +108,26 @@ export var RecordListView = Backbone.View.extend({
             const model = cell.getRow().getData().model;
             vreChannel.trigger('displayRecord', model);
         });
+        this.table.on("tableBuilt", this.connectToData.bind(this));
```
jgonggrijp: Is this binding really necessary?

frontend/vre/record/record.model.js
```diff
@@ -12,6 +13,16 @@ function get_fetch_url(record_uri) {
...
 }
 
+function getRecordTags(annotations) {
+    if (!annotations) return undefined;
+    var recordAnnotations = FilteredCollection(annotations, (annotation) => {
+        return !annotation.get('edpopcol:field') && annotation.get('motivation') === 'oa:tagging';
+    });
+    return recordAnnotations.map((annotation) => {
+        return annotation.getDisplayText();
+    }).join(", ");
```
jgonggrijp: I wrote `FilteredCollection` myself, so obviously I'm fond of it, but in this case, it is better not to use it. It incurs overhead relative to plain `_.filter` and in this case you're not getting anything back in return, because you discard the filtered collection again as soon as the `getRecordTags` method returns. The additional event bindings also have the potential to create a memory leak.

```suggestion
    var recordAnnotations = annotations.filter((annotation) => {
        return !annotation.get('edpopcol:field') && annotation.get('motivation') === 'oa:tagging';
    });
    return recordAnnotations.map((annotation) => {
        return annotation.getDisplayText();
    }).join(", ");
```

expressive minimalism variant:

```suggestion
    return annotations.chain()
    .filter((anno) => !anno.get('edpopcol:field') && anno.get('motivation') === 'oa:tagging')
    // TODO: replace by _.invoke when switching to Underscore
    .invokeMap('getDisplayText')
    .join(', ')
    .value();
```

In a situation where you really need the additional functionality of a full-blown collection with a short lifetime, you can prevent memory leaks by explicitly calling `.off().stopListening().reset()` just before it goes out of scope (or just before you `delete` it if you do that). For views it's `.off().stopListening().undelegateEvents().$el.remove()` (which is the default behavior of `Backbone.View#remove`) and for routers just `.off().stopListening()`.

frontend/vre/utils/tabulator-utils.js
```diff
@@ -171,6 +172,60 @@ const standardColumns = _.mapValues({
     );
 });
 
+/**
+ * To be used as the `valuesLookup` parameter of the `headerFilterParams`
+ * option of a Tabulator column definition. The return value is an array
+ * of filter options. The options we want to show are all unique values
+ * of the column, with support for cells that contain multiple values
+ * separated by a comma.
```
jgonggrijp, line 182: Great comment. 👍

frontend/vre/utils/tabulator-utils.js
```diff
@@ -171,6 +172,60 @@ const standardColumns = _.mapValues({
...
+ */
+function getUniqueValues(cell) {
+    var cells = cell.getColumn().getCells();
```
jgonggrijp, line 185: Does `getUniqueValues` get run for every row? Because then this line is introducing quadratic runtime complexity.

frontend/vre/utils/tabulator-utils.js
```diff
@@ -171,6 +172,60 @@ const standardColumns = _.mapValues({
...
+    var data = _.map(cells, c => c.getValue() && c.getValue().split(', '));
+    data = _.chain(data).flatten().filter(_.negate(_.isEmpty)).uniq().sortBy().value();
```
jgonggrijp: If I'm right, you're sorting strings, so there is no functional difference between stable and unstable sorting. If you're not using a custom comparison function, `sortBy` is going to incur needless overhead compared to `sort`. Also, `uniq` has quadratic runtime complexity over unsorted data but linear runtime complexity over sorted data, so it is better to sort first:

```suggestion
    // TODO: replace sortedUniq() by uniq(true) when switching to Underscore
    data = _.chain(data).flatten().filter(_.negate(_.isEmpty)).sort().sortedUniq().value();
```

frontend/vre/record/record.list.view.js
jgonggrijp: On second thought, maybe these lines should be in `initialize` rather than `render`. Or maybe just the first one. You don't want to repeat a line that creates an event binding if the emitter and the listener stay the same.

jgonggrijp (commented):
This looks OK to me, I just have a few comments on (technical) details.

> In the current implementation, all annotations are loaded one by one, which is inefficient. However, loading all annotations at once requires some design choices that I would like to discuss before starting work on them, and in the meantime I would like to work on some other issues. @jgonggrijp: if you think that we can better fix loading all annotations at once before merging this, I understand.

No. I agree this should be handled with a single request eventually, and I'm happy to discuss the design choices with you, but the PR can be merged with individual requests per record as far as I'm concerned.

frontend/vre/utils/tabulator-utils.js
tijmenbaarda: It only runs once whenever the dropdown is opened.

frontend/vre/record/record.list.view.js
tijmenbaarda: This works well! I have put the first line in `initialize`; the `invokeMap` line stays in `render()` because it has to be completed every time the records change. I just had to change `this.collection` by `this.collection.models` in the `invokeMap` call.

frontend/vre/record/record.list.view.js
tijmenbaarda: This line could be removed.

frontend/vre/record/record.model.js
tijmenbaarda: Done, changed by the expressive variant!

Issue #333, opened 2025-12-11
Make it clear how to show all available records, for catalogues where this is possible


tijmenbaarda:
Some catalogues allow showing all records by not entering any search words. This is not clear for the user, however.

PR #345, opened 2026-01-26
Maximize loading of additional records where applicable and prevent submitting empty queries

### LINKED ISSUE #298: ESTC cannot load 500 additional records
If you perform a query in the ESTC, this works and you can load 50 more records at a time. However, if you try to load 500 more records, this results in an error. After this, it is still possible to continue loading 50 more records at a time.

tijmenbaarda:
To be reviewed alongside with https://github.com/CentreForDigitalHumanities/edpop-explorer/pull/98.

This PR fixes #298 (loading 500 more records was not possible for ESTC and STCN), as well as #333 (grey out "Search" button if query field is empty, except if the reader supports this). Both issues required changes in the Explorer.

### DIFF
--- a/backend/catalogs/graphs.py
+++ b/backend/catalogs/graphs.py
@@ -153,7 +153,7 @@ class SearchGraphBuilder:
     def _get_partial_results(self) -> list[Record]:
         start = self._start
         end = min(self._end, self.reader.number_of_results)
-        results = [self.reader.records[x] for x in range(start, end)]
+        results = [self.reader.records[x] for x in range(start, end) if x in self.reader.records]
         return results  # type: ignore
 
     def _get_content_graph(self) -> Graph:
--- a/backend/catalogs/graphs.py
+++ b/backend/catalogs/graphs.py
@@ -153,6 +153,7 @@ class SearchGraphBuilder:
     def _get_partial_results(self) -> list[Record]:
         start = self._start
         end = min(self._end, self.reader.number_of_results)
+        # Get all applicable records from self.reader.records. Note that this is a dict with integers as keys
         results = [self.reader.records[x] for x in range(start, end) if x in self.reader.records]
         return results  # type: ignore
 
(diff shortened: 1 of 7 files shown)


backend/catalogs/graphs.py
```diff
@@ -153,7 +153,7 @@ def perform_fetch(self):
     def _get_partial_results(self) -> list[Record]:
         start = self._start
         end = min(self._end, self.reader.number_of_results)
-        results = [self.reader.records[x] for x in range(start, end)]
+        results = [self.reader.records[x] for x in range(start, end) if x in self.reader.records]
```
jgonggrijp, line 157: This confuses me. How could `x in self.reader.records` be `False`?

frontend/vre/search/search.view.js
```diff
@@ -8,18 +8,20 @@ export var SearchView = CompositeView.extend({
...
     events: {
         'submit': 'firstSearch',
+        'input #query-input': 'queryChanged',
     },
     subviews: [{
         view: 'alert',
         method: 'prepend',
     }],
     initialize: function() {
+        this.allowEmptyQuery = this.model.get('allowEmptyQuery');
```
jgonggrijp: You don't seem to be using this extra property on the view. You're using the model attribute directly and I think I also prefer that, so you could just remove this line.

jgonggrijp (approved):
Looks like a neat solution to me. I have a question and a comment, but no showstoppers in any case.

backend/catalogs/graphs.py
tijmenbaarda: This is because of the confusing fact that the `records` attribute is a `dict` with integers as keys... I will add a comment.

frontend/vre/search/search.view.js
tijmenbaarda: Oh, ok!

backend/catalogs/graphs.py
jgonggrijp: Then still, is the range of integer keys in that `dict` not contiguous?

PR #347, opened 2026-03-05
Upgrade backend and frontend dependencies

tijmenbaarda:
Conservatively upgrade backend and frontend dependencies. I checked the upgrades that could have breaking changes according to semver and I did some tests to check if everything is still working. Strangely enough, `rdflib` breaks from 7.3.0, so I pinned it to 7.2.1.

Soon we will have to do an upgrade of Django, because 4.2 is reaching end of life. I will request an upgrade of Python on the servers from 3.9 to 3.12, because Python 3.9 is not supported in newer Django versions.


jgonggrijp (approved):
Good work!

Issue #303, opened 2025-08-21
Deleting annotations does not always work (stateful)


jgonggrijp:
After creating a field annotation, you can click on it to edit it again. If you press the delete button, a confirmation popup appears. If you click "yes, really delete", the annotation *might* be deleted as intended, but something else might happen instead: the page scrolls back to top, the annotation editor and confirmation popup remain visible and the annotation is retained. Trying again usually does not help, but might work after refreshing the record.

Issue #239, opened 2024-12-11
Canonical sort no longer reflects actual fields from the record ontology


jgonggrijp:
The `FlatFields` collection type in `frontend/vre/field/field.model.js` still (since long ago) employs the `canonicalSort` function from `frontend/vre/utils/generic-functions.js` to determine the order of the fields. The code still works, but the array that defines the preferred order contains keys that no longer reflect the actual field names.

Maybe not important, but if we want to fix this, we need input from Jeroen Salman to re-establish the preferred order of the new keys.

PR #350, opened 2026-03-26
Assorted fixes

jgonggrijp:
This branch fixes three small issues, although the last one was quite tough to hunt down:

- The version numbers in the backend's `requirements-test.txt` had gone out of sync with the `requirements.txt`, which prevented the application from starting with `docker compose`.
- #239. I also changed the hardcoded field order to use a plain array, so that it is no longer necessary to explicitly manage numbers. I propose to let Jeroen edit this array directly so we can set the correct order once and for all.
- #303. An event handler that comes with Bootstrap's `modal` component out of the box was causing the modal to scroll up when we clicked on the deletion confirmation popover, which is technically outside of the modal. It took careful and patient debugging to figure this out, but fortunately the interference was easy to disable after that.


tijmenbaarda (approved):
Great!

Issue #351, opened 2026-04-09
Divergence between record ontology and VRE frontend


jgonggrijp:
The frontend has its own JSON-LD copy of the record ontology. Yesterday, I wanted to check whether it was up-to-date with the ontology in the dedicated repository, so I ran the script in the latter to generate its latest JSON-LD. The differences were more significant than I expected:

https://github.com/CentreForDigitalHumanities/EDPOP/compare/develop...bug/diverged-record-ontology

The red parts of the diff correspond to the VRE frontend, the green parts to the dedicated repository. I had to change the order of the output of the JSON-LD script in order to keep the diff minimal.

I will be investigating what needs to be done to reconcile the two. @tijmenbaarda if you could share anything you know that might help to clear up matters, please do.

jgonggrijp:
After looking through the Git commit history of both repositories, this gets a little bit more mysterious.

The dedicated ontology repo bumped the version to 0.1.0 in August 2024. After that, there was just one commit to add the `maximumRecordsPerPage` and `allowEmptyQuery` properties. Those are missing from the VRE repo; at least, that part is easy to explain and to reconcile.

The VRE repo has a commit from February 2025 (71bc8aacdc04e547e3e11c33be2a2fa3c5bb0d73) that appears to have done an integral update, in which the JSON-LD from the ontology repo was copied wholesale into the VRE repo. Evidence that the copy was wholesale includes that the order of the terms changes drastically, that a previously corrected typo was reintroduced and that other typos that appear in the ontology repo were also copied (e.g. `rdf:property` instead of `rdf:Property` (I will create an issue for that in the ontology repo)). However, it was not *just* a wholesale copy, because this commit also added properties that seem to have never existed in the dedicated ontology repo: `gender`, `holdings`, `genre`, `bibliographicalFormat`, `bookseller`, `typographicalFeatures`, `location`, `fingerprint`, `activityTimespan` and `collationFormula`.

After that, there have been two more commits to the copy of the ontology in the VRE repo:

- 0cffbb7 just changed a `prefLabel` and added a trailing newline;
- 18ec0e8 hand-edited the JSON-LD to add the `digitization` property. This explains why `digitization` is not present in the version of the ontology in the dedicated repo.

tijmenbaarda:
As discussed:
- The typos and other obvious mistakes should be corrected in the dedicated ontology repo
- All new fields can be added to the dedicated ontology repo, because otherwise they would disappear from the VRE
[...]

PR #352, opened 2026-04-09
Add annotations field to hardcoded copy of record ontology in frontend

jgonggrijp:
Parallel to https://github.com/CentreForDigitalHumanities/edpop-record-ontology/pull/10, I have copy-pasted the JSON-LD description of the new field to the hardcoded copy of the record ontology in the VRE frontend. This seemed the fastest way to ensure that Jeroen could start adding annotations, rather than going out of my way to fully address #351 first.

The only code change is a JSON-LD duplicate of https://github.com/CentreForDigitalHumanities/edpop-record-ontology/pull/10, so this is probably not interesting to review. However, please do think along with me: is this really enough to ensure that Jeroen can annotate? I think it is but I might be mistaken!

### DIFF
--- a/frontend/vre/edpop-record-ontology.json
+++ b/frontend/vre/edpop-record-ontology.json
@@ -704,6 +704,21 @@
       }
     },
     {
+      "@id": "edpoprec:annotations",
+      "@type": "rdf:Property",
+      "rdfs:domain": {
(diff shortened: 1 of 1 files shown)


tijmenbaarda (approved):
I am sorry for the late review! Yes, this works, and together with #353 this should be enough.

Issue #293, opened 2025-08-13
Annotations may disappear if records are reloaded


tijmenbaarda:
Field annotations may disappear from the VRE interface if records are reloaded from the Explorer in two cases:

* Annotations to fields that are missing from the original records disappear if the original record later includes these fields
* Annotations to fields that are present in the original record disappear if the values of the field change in the source database

The annotations will remain present in the triplestore and are also loaded to the frontend, but the frontend won't render them.

jgonggrijp:
I investigated this issue and I think the two subissues might have different causes.

I assigned myself on the assumption that I might be able to fix this today, but I think it is a bit too involved for this narrow time frame. I might still do it later but for now I relinquish it. My approach would to add a behavioral regression test for one scenario, step with the debugger until I understand what's going wrong, fix it and then repeat with the other scenario.

jgonggrijp:
Regarding the second subissue, #349 (when the field is originally present on the record), I am quite sure what is going on:

1. The `RecordDetailView` creates a `FlatterFields`, in which every `{field, value}` pair is represented as a separate model. This means there may be multiple models representing the same field. In addition, it includes `{field}` models for those fields that don't have a value in the originating catalog. The `FlatterFields` is passed to a `RecordFieldsView` instance.
2. The `RecordFieldsView` creates a `FieldView` for each of its models. In other words, there are one or more `FieldView`s per field, depending on the number of values for a field in the original record. Each `FieldView` also gets a collection of annotations matching the `{field, value}` model. This matching is based on the `annotationMatchesField` function.
3. Each `FieldView` then renders as one or more rows in the table of fields-with-annotations.

The disappearance happens in step 2. If an annotation has an `originalText`, and the record from the catalog no longer includes that text as a value of the annotated field, then `annotationMatchesField` returns `false`. Hence, the annotation in question is never passed to a `FieldView`.

This same mechanism does **not** explain the other subissue (#348, when the field was originally missing on the record). To the contrary: purely based on the logic above, I would actually expect annotations that lack an `originalText` to be _repeated_ in every `FieldView` with matching fieldname. I have not yet investigated why this is not the case. However, it might be worth postponing this until #349 has been fixed, because I think that will require a minor redesign.

[...]

PR #353, opened 2026-04-26
Allow corrections to fields and additions to empty fields

tijmenbaarda:
This PR allows making corrections to existing fields and additions to empty fields, instead of comments on fields. The corrections and additions are also shown in the table and are used for filtering and sorting.

I also made some alterations to make it possible to save annotations by pressing return and dismiss the annotation editor as well as the modal with the escape key.


frontend/vre/annotation/annotation.edit.view.js
```diff
@@ -5,16 +5,17 @@ import confirmDeletionTemplate from './annotation.confirm.deletion.mustache';
 import {glossary} from "../utils/glossary";
 
 export var AnnotationEditView = View.extend({
-    tagName: 'tr',
+    tagName: 'div',
```
jgonggrijp, line 8: Fun fact (no need to change): `div` is the default so you can also omit the `tagName` altogether.

frontend/vre/annotation/annotation.edit.view.js
```diff
@@ -38,6 +39,13 @@ export var AnnotationEditView = View.extend({
         if (this.model.get('motivation') === 'oa:tagging') {
             glossary.on('update', this.render);
         }
+        setTimeout(() => {
+            var el = this.$('textarea').get(0);
+            if (el) {
+                el.focus();
+                el.select();
+            }
+        }, 0);
```
jgonggrijp, line 49: I'm guessing you did a lot of tinkering before arriving at this solution. Maybe add a few comments on why you did it this way.

frontend/vre/field/field.model.js
```diff
@@ -11,15 +11,46 @@ import {
...
+        // Field missing in the source data: edpopcol:originalText should not be present
+        annotation = annotations.find(x => {
+            return x.get('edpopcol:field') === field.id && !x.get('edpopcol:originalText');
+        });
+    } else {
+        // Field present in the source data: check if edpopcol:originalText matches
+        annotation = annotations.find(x => {
+            return x.get('edpopcol:field') === field.id && x.get('edpopcol:originalText') === value['edpoprec:originalText']
+        });
+    }
```
jgonggrijp: Would be nice if you could get rid of the repetition here. A suggestion:

```suggestion
function annotationMatchesField(field, value) {
    var originalText = value && value['edpoprec:originalText'];
    return function(anno) {
        if (anno.get('edpopcol:field') !== field.id) return false;
        var annoText = anno.get('edpopcol:originalText');
        return originalText == null ? !annoText : annoText === originalText;
    };
}

/**
 * Get correction for a field value based on
 * @param {Object} value - the specific value to get the display of
 * @param {Field} field - the whole field object
 * @param {Annotations} [annotations] - annotations for the record
 * @returns {string|undefined}
 */
function getCorrectedTextForFieldValue(value, field, annotations) {
    if (!annotations) return undefined;
    var annotation = annotations.find(annotationMatchesField(field, value));
```

By the way, the `annotationMatchesField` looks familiar. Maybe we already have that function or something similar in another module.

frontend/vre/field/field.model.js
```diff
@@ -11,15 +11,46 @@ import {
...
+    if (annotation) {
+        return annotation.get('oa:hasBody');
+    }
+    return undefined;
```
jgonggrijp: This line is redundant.

frontend/vre/field/field.model.js
```diff
@@ -11,15 +11,46 @@ import {
...
+}
+
 /**
  * Get a default main display string of the `value` attribute of a
  * field flattened using {@link FlatterFields}. Currently, this is
  * the normalized "summary text" if available and otherwise the
- * original text from the source database.
- * @param {object} value
+ * original text from the source database. If
```
jgonggrijp: Is there something supposed to come after the "If"?

frontend/vre/field/field.model.js
```diff
@@ -28,16 +59,26 @@ export var Field = Backbone.Model.extend({
...
-        if (!value) return value;
-        if (_.isArray(value)) {
-            return _.map(value, getMainDisplayOfFieldValue).join(' ; ');
+        if (!value) {
+            // Field is missing in the source data: return annotation if present
+            var correctedText = getCorrectedTextForFieldValue(null, this, annotations);
+            if (correctedText)
+                return correctedText;
+            else
+                return '';
```
jgonggrijp:
```suggestion
            return getCorrectedTextForFieldValue(null, this, annotations) || '';
```

frontend/vre/field/field.model.js
```diff
@@ -28,16 +59,26 @@ export var Field = Backbone.Model.extend({
...
+        } else if (_.isArray(value)) {
+            // Field is repeated: concatenate all values
+            return _.map(value, (value) => getMainDisplayOfFieldValue(value, this, annotations)).join(' ; ');
+        } else {
+            return getMainDisplayOfFieldValue(value, this, annotations);
         }
```
jgonggrijp, line 75: You don't really need the `else`s here because every branch returns immediately. I won't insist that you need to remove them, though.

frontend/vre/field/field.view.js
```diff
@@ -23,12 +24,18 @@ export var FieldView = AnnotatableView.extend({
...
+    },
+
+    hasEdit: function() {
+        return this.collection.length > 0; // TODO check for just edits
     },
 
     renderContainer: function() {
         const templateData = {
             field: this.model.get('key'),
+            hasEdit: !this.hasEdit(),
```
jgonggrijp: Is the negation really meant to be there? You didn't mean to call the `templateData` key `hasNoEdit`?

frontend/vre/field/field.view.js
```diff
@@ -23,12 +24,18 @@ export var FieldView = AnnotatableView.extend({
...
+            isEmpty: typeof this.model.get('value') === 'undefined',
```
jgonggrijp: Models also have a `has` method: `this.model.has('value')` returns `false` if the requested attribute is unset, null or undefined.


```suggestion
            isEmpty: !this.model.has('value'),
```

frontend/vre/record/record.model.js
```diff
@@ -66,7 +65,10 @@ export var Record = JsonLdModel.extend({
             if (!this.isNew()) {
                 this.annotations.fetch();
             }
-            this.annotations.on('sync', () => this.trigger('annotations:loaded', this));
+            /* Trigger annotations:loaded to rerender the table row with annotations
+               on sync (which happens after fetching, creating and editing an annotation
+               and remove (which happens after deleting an annotation) */
+            this.annotations.on('sync remove', () => this.trigger('annotations:loaded', this));
```
jgonggrijp, line 71: How about using the `update` event? That covers both, as well as additions and changes.

jgonggrijp (changes requested):
Good work! I'm mostly happy with this.

~The one thing I find a bit counter-intuitive, when trying out the changes locally, is the way in which the pencil and addition buttons are placed. The addition button is only present on unset fields, while I would argue that it belongs on all fields, since a record might for example have multiple contributors. The pencil icon is only present on original text and not on corrections. The corrections can still be edited by clicking on them, but this is a difference with original text that I don't think users will naturally understand. They might get the impression that a correction is set in stone once made.~

~If this is tricky to address, you have my blessing to postpone it to a new issue so that Jeroen can get started annotating.~

*Edit: definitely postpone addressing the above point. See my recent comment in #293 on what I think should be done about this.*

Other than that, I have some code comments. Do with them what you want. The only thing I insist you address is the hanging "if".

frontend/vre/field/field.model.js
jgonggrijp: Found it. It's `annotationMatchesField` inside `field/record.fields.view.js`. However, that function might also be the source of #293. As far as I'm concerned you can reuse it, we just have to merge back the fix when I address #293.

jgonggrijp:
@tijmenbaarda I have done the following so I could already deploy this feature to production:

1. Create a `release/0.4.0` branch based off `develop` (which already included the annotations field).
2. Merge your `feature/corrections` into `release/0.4.0`.
3. Deploy `release/0.4.0` both to the acceptance server (as a sanity check) and to the production server.
4. Create draft release notes for version 0.4.0 (https://github.com/CentreForDigitalHumanities/EDPOP/releases/tag/untagged-8997d211614d5e74bfe0), mostly just to document the fact that a 0.4.0 release is already in preparation.

Once you finalize this branch, you can normalize the situation again with the following steps:

1. Merge this branch into `develop` as usual.
2. Merge `develop` into `release/0.4.0` (this might also bring in other changes that happened in the meanwhile).
3. Locally check that everything seems to work.
4. Deploy `release/0.4.0` to the acceptance server again, as a final sanity check.
5. Finish the `0.4.0` release as usual, creating a `0.4.0` tag on `master`.
6. On the production server, before deploying, login as `www-data`, navigate to `$WEBROOT/data/source` and run `git checkout master`. (If you skip this step, the deployment script will attempt to merge the `master` branch into the local `release/0.4.0` branch, which might fail.)
7. Switch the target branch back to `master` in the `dh_deploy_config.json` and deploy as usual.

frontend/vre/record/record.model.js
tijmenbaarda: Oh ok! I tried that now, but somehow it does not update when changing an annotation. Do you have any idea why that might be?

frontend/vre/field/field.view.js
tijmenbaarda: Oh, yes! I suppose I tried out the `!` and then forgot to change the key's name. Thanks!

jgonggrijp:
@tijmenbaarda The version prior to my review is already in production and you have since addressed my review. Do you agree this can be merged into develop?

If you do this, hereby a gentle reminder of https://github.com/CentreForDigitalHumanities/EDPOP/pull/353#issuecomment-4354677479.

tijmenbaarda:
Thanks - yes, just the frontend unit tests had to be fixed, which I finally did. I have now merged the pull requests and finished the release of 0.4.0!

PR #354, opened 2026-05-18
field multivalues

jgonggrijp:
This is a work-in-progress solution for #293 that builds on top of the changes in #353. #353 should be merged first, after which this branch will be rebased on develop. For now, this exists as a draft PR purely to inform you that the work is ongoing.


frontend/vre/field/field.model.js
```diff
@@ -152,9 +174,388 @@ export var FlatterFields = FlatFields.extend({
...
+ * @member {FilteredCollection<FlatterFields>} RecordFieldData#values
+ */
+/**
+ * Edits made on the current field in the current record.
+ * @member {FilteredCollection<Annotations>} RecordFieldData#annotations
+ */
+
+/**
+ * Presentation-oriented collection with the field values and edits for a single
+ * field of a given record.
```
tijmenbaarda: Maybe you can add here the name of the model type (`RecordFieldValueAttributes`) explicitly - the documentation of that model type is extremely helpful but it is a bit hard to find (my IDE doesn't really help at least :)). Maybe you can also explicitly clarify that this collection represents just one instance of a field, and not a combination of multiple repeated fields (is that correct?), and whether it may represent an empty field.

frontend/vre/field/field.model.js
```diff
@@ -152,9 +174,388 @@ export var FlatterFields = FlatFields.extend({
...
+});
+
+/**
+ * @typedef {Object} RecordFieldAttributes
+ * @property {module:'../utils/jsonld.model.js'.JsonLdModel} field
+ * @property {module:'../record/record.model.js'.Record} record
+ */
+
+/**
+ * Presentation-oriented model representing a specific field within a specific
```
tijmenbaarda: Here, I would have the same question as above.

frontend/vre/field/field.model.js
```diff
@@ -152,9 +174,388 @@ export var FlatterFields = FlatFields.extend({
...
+ * @param {Field} pair[0] - Original value.
+ * @param {Annotation} pair[1] - Edit.
+ * @returns {RecordFieldValueAttributes}
+ */
+function wrapCorrection(pair) {
+    const original = pair[0];
+    const edit = pair[1];
+    const originalText = originalTextSelector(edit);
+    if (edit.get('marksDeletion')) return {
+        id: originalText + ' →',
```
tijmenbaarda: I think I like the use of this arrow, but maybe it can be documented somewhere.

frontend/vre/field/field-value.view.js
```diff
@@ -0,0 +1,115 @@
...
+ */
+export var FieldValueView = View.extend(/**
+                                         * @lends FieldValueView.prototype
+                                         */{
+    tagName: 'tr',
+    template: fieldValueTemplate,
+    /** @member {module:'./field.model.js'.PresentableContents#model} model */
+
+    events: {
+        'click .fa-plus, .fa-pen': bubble('edit'),
```
tijmenbaarda: I like this use of `bubble` to keep consistency!

frontend/vre/field/field-value.view.js
```diff
@@ -0,0 +1,115 @@
...
+            edit = attributes.edit,
+            updatedDate = edit && edit.getUpdatedDate(),
+            originalText = attributes.originalText,
+            payload = _.extend({}, attributes, {
+                fieldInfo,
+                linkedRecordUri,
+                updatedDate,
+            });
+        if (edit && originalText && originalText.length > 20) {
+            payload.shortOriginalText = (
```
tijmenbaarda: Very nice that the long text is visible as a tooltip

tijmenbaarda (approved):
- "Deletions are not yet taken into account in table views": what I see is that [object Object] is rendered in this case. Also additions are not taken into account in table views, but I suppose that is equally easy to add. It is fine with me if we fix these problems in a separate PR.

frontend/vre/field/field.model.js
jgonggrijp: The connection between `CombinedFieldValues` and `RecordFieldValueAttributes` is documented in lines 372-379, due to the way in which JSDoc encodes this. However, my impression is that most editors don't implement JSDoc very well, so I'm not surprised that you're not seeing the information. I can add a `{@link}` to the description as a workaround.

To answer the other question, this collection represents all original values and all edits for a single field on a single record. Each model inside the collection does represent just one instance. I will elaborate on this in the description.

frontend/vre/field/field.model.js
jgonggrijp: The answer is also the same; it represents all values and edits for a single field in a single record. The reason we have both a model and a collection at this level of representation is that Backbone's models and collections have slightly different purposes. The collection is a member of the model. I will elaborate on this in the description.

frontend/vre/field/field.model.js
jgonggrijp: On lines 233-234, the documentation for `RecordFieldValueAttributes` mentions that the id is a composition of the original text and/or the corrected/added value. Would it be sufficient if I explicitly mention the arrow there?

